PR #1, opened 2019-01-31
IRAM_ATTR for the tick function

h3ndrik:
otherwise the esp32 will crash if cache is disabled (in my case during SPIFFS write).

Also add the attribute to your function:
    static void IRAM_ATTR lv_tick_task(void)

### DIFF
--- a/lv_conf.h
+++ b/lv_conf.h
@@ -6,6 +6,8 @@
 #ifndef LV_CONF_H
 #define LV_CONF_H
 
+#include "esp_attr.h"
+
 /*----------------
  * Dynamic memory
  *----------------*/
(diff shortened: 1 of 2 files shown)


kisvegabor:
Great! Thanks you!

kisvegabor:
I'm trying to use the display drivers with DMA but with no luck. If I set `past_cb` in the SPI driver it fails at the first sending. 

I this case lvgl is working with `LV_VDB_DOUBLE` to have 2 buffers, one for sending and other for rendering.

I can't find the root of the issue, although I'm not so experienced with ESP. I uploaded the current version to the [disp_drv_dma](https://github.com/littlevgl/esp32_ili9431/tree/disp_drv_dma) branch. Can you have a look at it?

kisvegabor:
I made it work. Now rendering and flushing work in parallel. This way I achieved > 30 FPS on a 320 x 240 display.

PR #10, opened 2019-07-11
Use Espressif recommended directory structure for ESP32 projects

### LINKED ISSUE #9: lvgl as a esp32-idf component
Hi,

Is it possible to add **lvgl** as a `component` (in the esp32 idf argon)? [here](https://github.com/C47D/esp32_lvgl_experiments) is an example on how (see components directory).

The **lvgl** directory should be added into a directory named `components`, the same goes for the **drv** directory. The `component.mk` file that goes into the **drv** directory may look like this example:

```
COMPONENT_SRCDIRS := .
COMPONENT_ADD_INCLUDEDIRS := $(COMPONENT_SRCDIRS) ..
```

The `component.mk` file that goes into the **lvgl** directory may look like this example:
```
CFLAGS += -DLV_CONF_INCLUDE_SIMPLE

COMPONENT_SRCDIRS := . \
	./src/lv_core \
	./src/lv_draw \
	./src/lv_objx \
	./src/lv_hal \
	./src/lv_misc \
	./src/lv_misc/lv_fonts \
	./src/lv_themes \
	./src/lv_fonts
COMPONENT_ADD_INCLUDEDIRS := $(COMPONENT_SRCDIRS) ..
```

C47D:
Fixes #9 and also updates the README. Please test it and let me know what do you think.

Carlos


embeddedt:
Now that I look at it, the two `*_component.mk` files are not very complicated. Wouldn't it make more sense to just keep them in the root folder of `lvgl`?

C47D:
@embeddedt Do you mean keeping the `lvgl_component.mk` on the main lvgl repo? Same for the `lv_example_component.mk` on the main lv_examples repo?

embeddedt:
Yes, IMO it would be cleaner than maintaining them in a separate repository.

C47D:
> Yes, IMO it would be cleaner than maintaining them in a separate repository.

Yes, we can try it.

I forgot to add:

```
ifndef LV_CONFIG_INCLUDE_SIMPLE
CFLAGS += LV_CONFIG_INCLUDE_SIMPLE
endif
```

on the top of the `lvgl_component.mk` file, i'm at work right now and can't add it until i arrive home.

C47D:
@embeddedt now that i think about it, the complexity of the make in esp32 based projects relies on the **esp-idf** library, it might be a better idea to keep the `component.mk` files just in this repo.

components/drv/ili9341.h
```diff
@@ -19,9 +19,9 @@ extern "C" {
  *      DEFINES
  *********************/
 #define DISP_BUF_SIZE (320 * 40)
-#define ILI9341_DC   19
-#define ILI9341_RST  18
-#define ILI9341_BCKL 23
+#define ILI9341_DC   2
+#define ILI9341_RST  4
+#define ILI9341_BCKL 21
```
kisvegabor: Why was it necessary to change the pins?

kisvegabor:
> Now that I look at it, the two *_component.mk files are not very complicated. Wouldn't it make more sense to just keep them in the root folder of lvgl?

Now there is no platform-specific things in `lvgl` and `lv_examples` repos. I'd concern about two things:
1. we will end up in a lot of platform-dependent config files which will make repos dirty
2. these config files might conflict (same name, one accidentally can use the other too)

components/drv/ili9341.h
C47D: I changed it to use my dev board, the one with only 30 gpios but forgot to change back to the pins you used. Do you want me to change them back?

kisvegabor: Altough, they are just randomly selected if there is no reason to change them please use the originals in this repo.

C47D: Ok, i will fix them once i arrive home

C47D:
@kisvegabor I think it's possible to change the included headers on the `main.c` file from:

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include "freertos/FreeRTOS.h"
#include "freertos/task.h"
#include "esp_system.h"
#include "driver/gpio.h"
#include "../lvgl/lvgl.h"
#include "../lv_examples/lv_apps/demo/demo.h"
#include "esp_freertos_hooks.h"


#include "../drv/disp_spi.h"
#include "../drv/ili9341.h"
#include "../drv/tp_spi.h"
#include "../drv/xpt2046.h"
```

to:

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

#include "freertos/FreeRTOS.h"
#include "freertos/task.h"

#include "esp_system.h"
#include "esp_freertos_hooks.h"

#include "driver/gpio.h"

#include "lvgl/lvgl.h"
#include "lv_examples/lv_apps/demo/demo.h"
#include "drv/disp_spi.h"
#include "drv/ili9341.h"
#include "drv/tp_spi.h"
#include "drv/xpt2046.h"
```

So there are no relative paths, what do you think?

C47D:
Never mind, i tried to implement [this](https://github.com/littlevgl/esp32_ili9341/pull/10#issuecomment-510883932) command and couldn't make it work.

I think this is ready to be merged :+1:

components/drv/ili9341.h
C47D: Fixed in 1140c49

kisvegabor:
@C47D 
I've needed to update include paths in the mk files. Does it seem reasonable to you?
However, it still didn't work because the Makeile next to main.c also should be updated with `LV_CONF_INCLUDE_SIMPLE`. (I haven't checked yet how should it be done)

C47D:
Is it? I'm already running it on the ESP-WROVER-KIT with no problems.

To add LV_CONF_SIMPLE_INCLUDE to the main.c just add CFLAGS+=-DLV_CONF_INCLUDE_SIMPLE to the component.mk of the main directory.

Let me know if it works.

kisvegabor:
It solved the issue. Thank you!

And the thank you this update too! :+1:

PR #25, opened 2019-10-29
Kconfig

### LINKED ISSUE #20: KConfig?
The standard convention in ESP-IDF libraries is to use KConfig / menuconfig / sdkconfig for user configuration of libraries/components. Right now this project uses a couple of makefile flags as well as header defines for configuration of the library for the end-user's device. I would suggest a move to use all header defines but have those defines in turn reference sdkconfig variables. 

So in the header for example we'd change this:

```c
#define DISP_SPI_MOSI 23
```

to 

```c
#define DISP_SPI_MOSI CONFIG_DISP_SPI_MOSI
```

`CONFIG_DISP_SPI_MOSI` in turn is set in sdkconfig directly or through menuconfig, and will still have a default value. The two layers of defines gives you the flexibility to quickly change it in the library header while testing, which does not recompile the whole sdk.

The advantage for the user is this similar to other libraries as well as all the IDF components themselves, but also that they do not necessarily have to vendor the library, they can keep it as a git submodule without worrying about the dirty flags.

I very recently [implemented such a scheme](https://github.com/jeremyjh/ESP32_TFT_library/blob/master/components/tft/Kconfig) for the loboris driver and would be happy to submit a PR for this library as well.

jeremyjh:
resolves #20 

Implements suggestions from #20 to use KConfig for end-user configuration. It also includes support and instructions for CMake (when using IDF 4.x). Part of the reason for wanting the KConfig changes is so that this repository can be used as a library submodule without copying or changing any files and thus keeping the submodule status clean. This makes updates easier and encourages reproducible builds. There are other minor changes to build files and paths to support this goal. A good way to get an overview is to look at the new instructions in the [rendered README](https://github.com/jeremyjh/esp32_ili9341/blob/kconfig/README.md).

I've tested the best I can, which is not terribly great because the only ili9341 board I have is the M5 which has no touch controller. So the method I used to test all the different configurations was to set them one at a time and (painstakingly) compare the pins in `sdkconfig` to what they were in the old README for each board. I'd feel better if someone can test at least one configuration that includes a touch controller though.


kisvegabor:
@jeremyjh 
Thank you!

@C47D 
Can you take look at these changes?

C47D:
Hi,

Yes, I can take a look at them, I wasn't familiar with this capabilities but they look great!

C47D:
Here are some screenshots :)

After `idf.py menuconfig`
imagen

imagen

imagen

imagen

imagen

imagen

imagen

imagen

I found some typos (third screenshot is says **predfined** instead of **predefined**), but overall is a big improvement on the user experience, i really liked the predefined display configurations, they save a lot of time.

After saving the configuration i had some troubles to build the project.
- In `components/lvgl_ili934/ili9341.h` I had to change `#include "lvgl/lvgl.h"` to `#include "lvgl.h"`
- Same in `components/lvgl_ili934/disp_spi.c`
- Same in `components/lvgl_ili9341/xpt2046.h`
- Same in `main.c`

Maybe i missed some steps, can you reproduce my issue @jeremyjh? I choosed the Wrover Kit v4.1 default configuration.

jeremyjh:
@C47D I am able to reproduce a build error with make so I should be able to fix it. I think its not including sdkconfig.h in the translation unit compiling lv_conf. I think the last time I tested make after adding the vars in lv_conf  I hadn't switched my IDF version; it builds on 4.0 with make but not on 3.x. Thanks for the screenshots I will add them to the README too.

C47D:
Hi @jeremyjh,

I will try to test it again (at work), I don't remember which version of the IDF I used yesterday.

> Thanks for the screenshots I will add them to the README too.

No problem, I would like to take a video of the project setup but sometimes it's better to have it written down.

Let me know if you think I can help with more tests.

C47D:
I played a bit with the Kconfig and added an option to let the user choose if they want to control the display backlight with an GPIO in case they want to have it always enabled by using a wire and not an GPIO.

jeremyjh:
I pushed a commit that fixed the typos and also includes the `sdkconfig.h` in lv_conf.h; I'm now able to build with make on IDF 3.x. I pushed another with your patch to enable backlight control. I also added check to the init function to not to send a value to the BCKL pin in case this option is not enabled.

@C47D I think the include of sdkconfig.h was the issue you ran into; you shouldn't need to use the CONFIG_ defines inside source files, the former names are all still defined in headers; for example in tp_spi.h there is:

`#define ENABLE_TOUCH_INPUT  CONFIG_LVGL_ENABLE_TOUCH
`
in ili9341.h there is:

```c
#if CONFIG_LVGL_BACKLIGHT_ACTIVE_LVL
  #define ILI9341_BCKL_ACTIVE_LVL 1
#else
  #define ILI9341_BCKL_ACTIVE_LVL 0
#endif
```

One awkward thing with bools in Kconfig is when the value is false, it is not defined at all, so we have to do above dance in case we actually use the 0 value and not just check for its definition in an `#if`

In other cases I still have two-layer of defines because it makes it faster to debug if you can just edit the header define; one drawback with KConfig is any change to sdkconfig forces a re-build of the whole SDK and all other libraries used in the project.

C47D:
> @C47D I think the include of sdkconfig.h was the issue you ran into; you shouldn't need to use the CONFIG_ defines inside source files, the former names are all still defined in headers; for example in tp_spi.h there is:
> 
> #define ENABLE_TOUCH_INPUT CONFIG_LVGL_ENABLE_TOUCH
> in ili9341.h there is:
> 
> #if CONFIG_LVGL_BACKLIGHT_ACTIVE_LVL
>   #define ILI9341_BCKL_ACTIVE_LVL 1
> #else
>   #define ILI9341_BCKL_ACTIVE_LVL 0
> #endif

You're right @jeremyjh, it should be as below, isn't?
```c
#if ENABLE_TOUCH_INPUT
#define ILI9341_BCKL_ACTIVE_LVL 1
#else
#define ILI9341_BCKL_ACTIVE_LVL 0
#endif
```

> One awkward thing with bools in Kconfig is when the value is false, it is not defined at all, so we have to do above dance in case we actually use the 0 value and not just check for its definition in an #if
> 
> In other cases I still have two-layer of defines because it makes it faster to debug if you can just edit the header define; one drawback with KConfig is any change to sdkconfig forces a re-build of the whole SDK and all other libraries used in the project.

I wasn't familiar with Kconfig syntax until yesterday, there must be a way to set a value to false and have the symbol defined.

So, are you now able to build the project with both IDF 3.0 and 4.0? If so i think this is ready to merge and see it used on the field so we can fix any missing errors the users found.

embeddedt:
> I wasn't familiar with Kconfig syntax until yesterday, there must be a way to set a value to false and have the symbol defined.

Not that I know of, because when you set a symbol to false, this happens:

```
CONFIG_TRUE_SYMBOL=y
# CONFIG_FALSE_SYMBOL is not defined
```

The only solution I can see is to modify the Kconfig to C header conversion algorithm to output defines with a value of 0 (which is out of scope here).

jeremyjh:
@C47D It does build, flash and run for me on both 4.0 and 3.x. 

After rebasing on master I found 4.0 compile (using GCC 8.2) was broken due to warning:

```
./components/lvgl_ili9341/ili9341.c:156:5: error: this 'while' clause does not guard... [-Werror=misleading-indentation]
     while(disp_spi_is_busy());
     ^~~~~
../components/lvgl_ili9341/ili9341.c:157:4: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'while'
```

It could be fixed by just adding empty braces e.g. `while(disp_spi_is_busy()) {} `

But I went ahead and [implemented](4c78b82) @embeddedt [suggestion](https://github.com/littlevgl/esp32_ili9341/commit/a0c6db7ff9dbbdffa4a186c44a3885a21dde5bc1#r35727494) to yield here.

C47D:
@jeremyjh Great news! If everybody agree I think it's ready to merge. Maybe on the future we can choose which of the example projects to build using menuconfig.

kisvegabor:
@jeremyjh @C47D 
Thank you very much for testing. 
It's awesome that it's working now with IDF3 and 4 too!

PR #27, opened 2019-11-03
[Kconfig] Fix the backlight active level on Wrover kit and component.mk cleanup

C47D:
The backlight in the Wrover kit v4.1 is active low.


kisvegabor:
Thank you!

PR #29, opened 2019-11-06
[Kconfig] Add basic ranges to valid GPIO numbers

C47D:
So the user doesn't choose non existent gpio pin numbers. This doesn't validate if the selected pins are in fact able to do MOSI, SCK, etc, it just checks that the selected pin is not 40 or larger (esp32 only have up to 39 pins).


kisvegabor:
Thank you!

PR #30, opened 2019-11-06
[disp_spi] Add taskYIELD to the busy loops and remove unnecessary inc…

C47D:
…lude

### DIFF
--- a/components/lvgl_ili9341/disp_spi.c
+++ b/components/lvgl_ili9341/disp_spi.c
@@ -10,10 +10,11 @@
 #include "esp_system.h"
 #include "driver/gpio.h"
 #include "driver/spi_master.h"
-#include <string.h>
+
 #include <freertos/FreeRTOS.h>
 #include <freertos/semphr.h>
-#include "freertos/task.h"
+#include <freertos/task.h>
+
 #include "lvgl/lvgl.h"
 #include "ili9341.h"
 
@@ -81,7 +82,9 @@ void disp_spi_send_data(uint8_t * data, uint16_t length)
 {
     if (length == 0) return;           //no need to send anything
 
-    while(spi_trans_in_progress);
+    while(spi_trans_in_progress) {
+	taskYIELD();
+    }
 
     spi_transaction_t t = {
         .length = length * 8, // transaction length is in bits
@@ -98,7 +101,9 @@ void disp_spi_send_colors(uint8_t * data, uint16_t length)
 {
     if (length == 0) return;           //no need to send anything
 
-    while(spi_trans_in_progress);
+    while(spi_trans_in_progress) {
+	taskYIELD();
+    }
 
     spi_transaction_t t = {
         .length = length * 8, // transaction length is in bits


kisvegabor:
We had a [discussion](https://github.com/littlevgl/lv_port_esp32_ili9341/commit/a0c6db7ff9dbbdffa4a186c44a3885a21dde5bc1#r35777238) about `yield` vs `bocking`. I voted for the blocking wait because of performance issues.

This is a more visible place so let's continue the discussion here.

cc @jeremyjh @embeddedt

C47D:
Hi, I saw the discussion and implemented the same solution used on here:
https://github.com/littlevgl/lv_port_esp32_ili9341/blob/ac2102c35972a0dfbc9bb44ccfa553b4d23650bd/components/lvgl_ili9341/ili9341.c#L176-L183

Is there a way to measure and decide which `yield` or `blocking` is better?

embeddedt:
> Is there a way to measure and decide which yield or blocking is better?

You would have to get a timestamp before and after the `while` loop in both cases. Then you could compare the delay.

C47D:
From https://github.com/littlevgl/lv_port_esp32_ili9341/commit/a0c6db7ff9dbbdffa4a186c44a3885a21dde5bc1#r35762687

> According to [the docs](https://www.freertos.org/a00020.html#taskYIELD) taskYIELD is not needed if FreeRTOS is in preemptive mode, which it is in ESP32 by default. But, that only refers to running another task at higher priority. If the other task is same priority, even if it is not starved it will have to wait on useless busy-wait cycles. I guess if the user finds this to be a problem, they can run the LVGL task at a lower priority.
> 
> Still, we shouldn't be waiting anywhere near 1000 us for a taskYIELD that finds no other task to run. Maybe we should measure it before we decide; I probably won't have time today but can do it this weekend. If another task at the same priority is runnable, it _should_ be run while we are waiting on I/O. This is the basic principle of async I/O; its better to let the CPU do other things than busy wait.

We could do a couple of experiments:
- Running the demo without any other tasks, measure both blocking and yield approaches.
- Running the demo alongside other tasks, higher priority and maybe lower priority, and again measure the blocking and yield approaches.

I haven't worked yet with the wifi or ble esp32 capabilities but I guess those are handled by underlying tasks, and we could also use those peripherals because it's most likely the esp32 lvgl users will use them.

How can we design the experiment?

embeddedt:
> How can we design the experiment?

This is my idea of how it might work.


We'd need to have some type of work running on another thread. We can either use a real peripheral like you are suggesting, or create some fake work by having some other thread(s) busy wait for a few milliseconds and then sleep.

Then we would add a timestamp inside the `ili9341_send_data` function:

```c
static void ili9341_send_data(void * data, uint16_t length) 
{ 
          uint32_t start_ms, end_ms;
          start_ms = sys_ticks(); /* get system uptime in milliseconds */
 	  while(disp_spi_is_busy()) { 
 			taskYIELD(); 
 	  } 
          end_ms = sys_ticks(); /* get system uptime in milliseconds */
          printf("Time to send: %u ms\n", end_ms - start_ms);
 	  gpio_set_level(ILI9341_DC, 1);	 /*Data mode*/ 
 	  disp_spi_send_data(data, length); 
 } 
```

We could comment out or include `taskYIELD` as necessary. We might need a high-precision timer rather than relying on the OS tick mechanism depending on the processor's speed. I used `sys_ticks` mainly as an illustration of how it could work.

I think that would give a good indication of how much performance is increased/decreased by yielding/not yielding.

C47D:
Do you think this (High Resolution Timer)[https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/system/esp_timer.html] can work or maybe this (xTaskGetTickCount)[https://www.freertos.org/a00021.html#xTaskGetTickCount] other function can help us.

The hardware timer resolution is in microseconds.

jeremyjh:
Hey guys. Sorry I've been quiet on this the last many days I've been traveling for work and really busy, I'd wanted to do a test along these lines but, actually I sort of wish I'd never opened that comment thread because in truth I misunderstood the scope of I/O operations in flush. I thought we spun while waiting on an entire DISP_BUF of data, but in fact as @kisvegabor points out each call to flush results in only 4 bytes sent in two operations. Busy wait on 4 bytes serial I/O doesn't seem excessive.

C47D:
Hi,

I have tested this PR in the demo example and haven't seen any issue. Actually `vTaskDelay` calls `taskYIELD` internally, i guess we're doing premature optimization.

kisvegabor:
@jeremyjh 
>  I sort of wish I'd never opened that comment thread

I haven't examined this case so deeply so it's just pure luck that it works. :slightly_smiling_face:  

I still believe any task yields will be too long (in such a performance-critical place) compared to sending a  few bytes. Imagine that FreeRTOS has 5ms heartbeat and lvgl has 1/10 screen-sized buffer. If FreeRTOS lets another task to run for only 1-time slice then it adds 5*10=50ms time to load a screen.

C47D:
We still need to measure things to know which method (blocking or yielding) have a bigger impact on performance.

@embeddedt Using a digital output instead of printing the timestamp can help? Something like this:
```c
static void ili9341_send_data(void * data, uint16_t length) 
{ 
          uint32_t start_ms, end_ms;
          
          gpio_set_level(DEBUG_PIN, 1);
 	  while(disp_spi_is_busy()) { 
 			taskYIELD(); 
 	  } 
          gpio_set_level(DEBUG_PIN, 0);

 	  gpio_set_level(ILI9341_DC, 1);	 /*Data mode*/ 
 	  disp_spi_send_data(data, length); 
 } 
```

Then we could see the gpio signal on a digital oscilloscope or using a logic analyser.

embeddedt:
@C47D Yes; that method would work as well.

> Imagine that FreeRTOS has 5ms heartbeat and lvgl has 1/10 screen-sized buffer. If FreeRTOS lets another task to run for only 1-time slice then it adds 5*10=50ms time to load a screen.

I think it depends on the priority of the other task relative to ours. If we have a higher priority we should be able to take back control before the timeslice of the other task ends. If we have the same priority then the other task will get the full timeslice (unless it sleeps beforehand).

C47D:
> @C47D Yes; that method would work as well.

I added the code to toggle an gpio and the display refresh went really slow, i didn't even had to measure the pulse width to know the performance drop. I will try again this week.

jeremyjh:
@embeddedt The [freeRTOS doc](https://www.freertos.org/a00020.html#taskYIELD) suggests taskYIELD is unnecessary to switch to a higher priority task as the highest priorty runnable task will always be running when `configUSE_PREEMPTION` is enabled (which it is in IDF). It may mean more frequent context switches between tasks with the same priority.

```
If configUSE_PREEMPTION is set to 1 then the RTOS scheduler will always be running the highest priority task that is able to run, so calling taskYIELD() will never result in a switch to a higher priority task.
```

C47D:
So we end up on the same task, calling taskYIELD only cost us a context switch?

PR #43, opened 2019-12-19
extended driver ili9341 to support dislplay rotation

rangerjo:
Allows rotating the display in all 4 directions. Uses the rotated feature of the lvgl_v6 driver struct and the different scanning directions of the ili9341 driver ic.

### DIFF
--- a/components/lvgl_ili9341/ili9341.c
+++ b/components/lvgl_ili9341/ili9341.c
@@ -15,6 +15,10 @@
 /*********************
  *      DEFINES
  *********************/
+#define ILI9341_MAD_MY  0x80
+#define ILI9341_MAD_MX  0x40
+#define ILI9341_MAD_MV  0x20
+#define ILI9341_MAD_BGR 0x08
 
 /**********************
  *      TYPEDEFS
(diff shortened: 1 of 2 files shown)


components/lvgl_ili9341/ili9341.c
```diff
@@ -159,6 +166,37 @@ void ili9341_enable_backlight(bool backlight)
...
+/**
+ * Rotate the display by 0, 90, 180 or 270 degrees.
+ * @param driver pointer to lvgl driver variable
+ * @param direction 0: 0, 1: 90, 2: 180, 3: 270 degrees
+ */
+void ili9341_rotate(lv_disp_drv_t * drv, uint8_t direction) {
+	uint8_t data[1];
+	switch(direction) {
+		/* 0 degrees: no rotation */
+		case 0: data[0] = (ILI9341_MAD_MV | ILI9341_MAD_BGR);
```
jeremyjh, line 178: These command needs to be different if `ILI9341_INVERT_DISPLAY` is enabled. I think you can OR with a `#define TFT_INVON   0x21`. You should be able to see that invert objects on the display even if your device looks correct with out it.

rangerjo: Just left my office for xmas holidays, I will catch up on this in the new year. From your response I'm reading you would be willing to merge this feature into the master branch, but want it to work for the inverted display option too? I'll give it a try!

PR #44, opened 2019-12-24
[WORK IN PROGRESS] Add support for ILI9488 display controller

### LINKED ISSUE #42: Support ILI9488
Is it possible to support the IL9488? As far as I understand, the differences are quite minimal. Thanks

C47D:
In order to support ILI9488 TFT display controller, fix #42


C47D:
Hi,
Sorry for the huge delay, this is still work in progress but i would like to have a review of the code done so far. I have tested the changes with my Wrover kit v4 and it still works, but i don't have any display with the ili9488 driver.

@jeremyjh Hi, do you think is a good idea to choose the right driver option when selecting predefined kits? Is it possible?
Also, i would like to rename the lvgl_ili9341 directory to something else and maybe have the ili9488 and ili9341 files into a directory named `drivers`, similar to what you've done on your other project. What do you think?

@rdelcorro Hi, `ili9488_init` is just a copy of the `ili9341_init` function, do you happen to have a working init function for your driver?

rdelcorro:
@C47D I do have an initialization sequence that can be used here. Let me know if I should create a branch / PR to push it to this branch or you would like to port it. Its here: https://github.com/Bodmer/TFT_eSPI/blob/master/TFT_Drivers/ILI9488_Init.h

I also have a fully working example of littlevgl with this driver but under the hood, its using the raw drawPixel of TFT_eSPI which is extremely slow. It renders a frame every 2 seconds or so. https://github.com/rdelcorro/TFT_espi_littlegl

C47D:
Thanks for the information @rdelcorro, I will take a look at it and try to adapt it, @jeremyjh also have a init sequence (I just found out)

https://github.com/jeremyjh/ESP32_TFT_library/blob/cf347104a635a876fe9614ff465d5b9dec10f255/components/tft/tftspi.h#L426-L490

Let me try to update the init sequence, and then you can update it to fit your needs, I also need to found out how to assign the display driver to the predefined kits and this PR will be ready.

C47D:
@rdelcorro Can you test this PR? I updated the ILI9488 init sequence based on your link and the other link, here's the menuconfig LVGL configuration, on predefined display you should choose none and all the options will become available...

imagen

rdelcorro:
@C47D tested and the menu seems to be working ok but the display is not looking good: https://photos.app.goo.gl/ATkH9zy2NGKBpAeh6. Let me know if I should try other settings

C47D:
Hi,

Are you editing the init sequence of the display?

rdelcorro:
I am not touching the code but rather just use the menuconfig, set the proper GPIO and then compile / flash

rdelcorro:
I just created a discord channel here if you like to join: https://discord.gg/2XsTzYt

C47D:
@rdelcorro I've updated a couple of register configurations on the init sequence (based on [this](https://github.com/jaretburkett/ILI9488/blob/master/ILI9488.cpp) configuration), let me know if that configuration works better. The ILI9488 datasheet is massive!

I've also accepted the Discord invitation and sent a message on the channel, contact me wherever is easier for you. Happy new year.

rdelcorro:
Hi @C47D, I am seeing the same artifacts with the new changes. I will try to debug a bit and report back. Happy new year!

C47D:
Sorry for taking so long to solve this, i haven't got the display i buyed from ebay to test before pushing code. Hopefully it works now.

OFF TOPIC:
Just realized i was using my work user name and email on the latest commits.

kisvegabor:
> Sorry for taking so long to solve this, i haven't got the display i buyed from ebay to test before pushing code. Hopefully it works now.

That's fine. We are very grateful for you and @rdelcorro for adding this! :slightly_smiling_face: 


> Just realized i was using my work user name and email on the latest commits.

No problem, if it's ok for you too.

Do you think we should rename the project to `lv_port_esp32` or `lv_port_esp32_ili`?

C47D:
@kisvegabor Yes, actually I was thinking on suggesting the name change, I would also like to have @jeremyjh opinion on that.

jeremyjh:
I would suggest just change it to `lv_port_esp32`. Even the ST* (ST7735 etc) family is not so different so if someone wanted to use that with LVGL it would make sense to add to this project as well.

kisvegabor:
> I would suggest just change it to lv_port_esp32. Even the ST* (ST7735 etc) family is not so different so if someone wanted to use that on ESP32 with LVGL it would make sense to add to this project as well.

Sounds reasonable.

Is this PR ready to merge?

C47D:
I'm only waiting somebody to validate it, I implemented the changes @Mat-Alm commented [here](https://github.com/littlevgl/lv_port_esp32_ili9341/issues/42#issuecomment-577772669), but I don't have the hardware to test it.

kisvegabor:
@Mat-Alm could you try the latest version?

components/lvgl_ili9341/ili9488.c
```diff
@@ -125,13 +126,13 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
     }
 #else
     lv_color16_t *buffer_16bit = (lv_color16_t *) color_map;
-    uint8_t mybuf[3 * size * sizeof uint8_t];
+    uint8_t mybuf[3 * size * sizeof(uint8_t)];
```
Mat-Alm: Change to uint8_t* mybuf = (uint8_t*) heap_caps_malloc(3 * size * sizeof(uint8_t), MALLOC_CAP_DMA);

or it will have stack overflow

components/lvgl_ili9341/ili9488.c
```diff
@@ -168,7 +169,7 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
 	/*Memory write*/
 	ili9488_send_cmd(ILI9488_CMD_MEMORY_WRITE);
 
-#ifdef 24BIT_DATA
+#if defined DATA_24BITS
 	ili9488_send_color((void*) buffer_24bit, size * 3);
 #else
 	ili9488_send_color((void*) buffer_16bit, size * 3);
```
Mat-Alm: Change to:

	ili9488_send_color((void*) mybuf, size * 3);
	heap_caps_free(mybuf);

components/lvgl_ili9341/ili9488.c
```diff
@@ -113,8 +113,9 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
 {
     uint32_t size = lv_area_get_width(area) * lv_area_get_height(area);
 
-// #define 24BIT_DATA
-#ifdef 24BIT_DATA
+#define DATA_24BITS
+
+#if defined DATA_24BITS
```
Mat-Alm: The code associated with DATA_24BITS does not work please remove it

Mat-Alm: Only RGB 565 to RGB 666 works

components/lvgl_ili9341/ili9488.c
```diff
@@ -168,7 +169,7 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
 	/*Memory write*/
 	ili9488_send_cmd(ILI9488_CMD_MEMORY_WRITE);
 
-#ifdef 24BIT_DATA
+#if defined DATA_24BITS
 	ili9488_send_color((void*) buffer_24bit, size * 3);
```
Mat-Alm: Does not work

Mat-Alm:
I have tested the version (I made comments on parts that fails) ... The part added related with #if DATA_24BITS  conversion does not work!  Only the RGB565 to RGB666 works like I previous stated!

it's possible to add custom pin setup on idf menu? Also based on the display choice is possible to auto set the resolution?

Rotation  function is not tested also better disable it. On initialization  it already put the screen on landscape mode.

Mat-Alm:
Also set in  lv_conf.h to 16bit or you will overflow ram section if you select RGB8888 Maybe if you put some buffer on devices with spiram it might work

C47D:
@Mat-Alm Thanks for trying this out, I will update the code with your comments as soon as I arrive home later today.

components/lvgl_ili9341/ili9488.c
C47D: Addressed in e8224db

components/lvgl_ili9341/ili9488.c
C47D: Addressed in e8224db

components/lvgl_ili9341/ili9488.c
C47D: Addressed with ef6be68, one question tho, do we need to free the allocated memory on the flush function or when the DMA engine is done transferring it?

components/lvgl_ili9341/ili9488.c
C47D: Addressed with ef6be68

C47D:
> it's possible to add custom pin setup on idf menu?

When selecting None on the Predefined board pinout option you should be able to select different pin numbers for your signals. Or you mean something else?

> Also based on the display choice is possible to auto set the resolution?

Yes, it is possible, maybe we can add it on another PR, once @kisvegabor change the repo name, and I also wanted to change some directory names to try to help add more drivers more easily.

> Rotation function is not tested also better disable it. On initialization it already put the screen on landscape mode.

Which rotation function? Sorry but this PR is kind of old and I tend to forget code I don't usually work with.

> Also set in lv_conf.h to (RGB565)16bit display output or you will overflow ram section if you select RGB8888. Maybe if you put some buffer on spiram it might work.

It already is:

https://github.com/littlevgl/lv_port_esp32_ili9341/blob/edf5bf19083e911876318c0842c091a6dfdc6d24/components/lvgl/lv_conf.h#L28-L34

Or do you mean somewhere else?

Mat-Alm:
> 
> 
> > it's possible to add custom pin setup on idf menu?
> 
> When selecting None on the Predefined board pinout option you should be able to select different pin numbers for your signals. Or you mean something else?
> 
For me it does not show any options for pin choice... 


idf

embeddedt:
You also have to choose "None" for the "predefined board pinout", as far as I can tell.

C47D:
@Mat-Alm I get it now, as your board doesn't have the display hardwired to the ESP32 you should be able to choose what pins to use, you can use the @embeddedt suggestion while we try to fix this issue.

Mat-Alm:
> Which rotation function? Sorry but this PR is kind of old and I tend to forget code I don't usually work with.
> 
Sorry it's not rotation but invert function... This piece of code at ili9488.c 

#if ILI9488_INVERT_DISPLAY
	uint8_t data[] = {0x68};
	// this same command also sets rotation (portrait/landscape) and inverts colors.
	// https://gist.github.com/motters/38a26a66020f674b6389063932048e4c#file-ili9844_defines-h-L24
	ili9488_send_cmd(0x36);
	ili9488_send_data(&data, 1);
#endif

Mat-Alm:
> 
> 
> @Mat-Alm I get it now, as your board doesn't have the display hardwired to the ESP32 you should be able to choose what pins to use, you can use the @embeddedt suggestion while we try to fix this issue.
> 
> I've created some issues on the usage problems you have found so we can work on them separately.
> 
>     * Choosing ESP32 DevKit v4 (38 GPIOS) as board doesn't allow the user to set custom pins for display control (#46)

Yes you got the point!

Mat-Alm:
> It already is:
> 
> https://github.com/littlevgl/lv_port_esp32_ili9341/blob/edf5bf19083e911876318c0842c091a6dfdc6d24/components/lvgl/lv_conf.h#L28-L34
> 
> Or do you mean somewhere else?
Sorry I had written what I thought wrong.

Well the display drivers only work on RGB565 mode right... If the user choose another mode it does not show any warning and it will make the system crash. So it's better to make some mechanism at compile/run time to detect it.

Mat-Alm:
Two more things:

At lv_conf.h the field:
#define LV_COLOR_16_SWAP   1
should be 
#define LV_COLOR_16_SWAP   0

At ili9488.c  the line with:

gpio_set_direction(ILI9488_BCKL, GPIO_MODE_OUTPUT);

Should be conditioned to the use of backlight pin control option on idf menu settings or it can cause problems.

C47D:
> 
> 
> > @Mat-Alm I get it now, as your board doesn't have the display hardwired to the ESP32 you should be able to choose what pins to use, you can use the @embeddedt suggestion while we try to fix this issue.
> > I've created some issues on the usage problems you have found so we can work on them separately.
> > ```
> > * Choosing ESP32 DevKit v4 (38 GPIOS) as board doesn't allow the user to set custom pins for display control (#46)
> > ```
> 
> Yes you got the point!

Lets continue this on the other issue.

C47D:
> 
> 
> > It already is:
> > https://github.com/littlevgl/lv_port_esp32_ili9341/blob/edf5bf19083e911876318c0842c091a6dfdc6d24/components/lvgl/lv_conf.h#L28-L34
> > 
> > Or do you mean somewhere else?
> > Sorry I had written what I thought wrong.
> 
> Well the display drivers only work on "RGB565-->RGB666" mode right? If the user choose another mode it does not show any warning and it will make the system crash. So it's better to make some mechanism at compile/run time to detect it.

I will create another issue to solve this.

C47D:
> 
> 
> Two more things:
> 
> At lv_conf.h the field:
> #define LV_COLOR_16_SWAP 1
> should be
> #define LV_COLOR_16_SWAP 0
> 
> At ili9488.c the line with:
> 
> gpio_set_direction(ILI9488_BCKL, GPIO_MODE_OUTPUT);
> 
> Should be conditioned to the use of backlight pin control option on idf menu settings or it can cause problems.
> 
> I had tested with the last version with this two more fixes and it works ok!

I've updated the code with this fixes, please let me know if it works now

Mat-Alm:
All working fine now!

C47D:
@kisvegabor ready to merge

kisvegabor:
Thank you guys! :rocket: 
Merged!

PR #45, opened 2020-01-17
Fixed: GPIO15 not working as Blacklight, set permanent to high

mcrobs:


mcrobs:
This is a necessary fix to use this case with an esp32
https://www.hwhardsoft.de/deutsch/projekte/arduitouch-esp/

jeremyjh:
Did you first try just setting `CONFIG_LVGL_BACKLIGHT_ACTIVE_LVL=y`? If that won't work I think it would be better to add a specific config. I don't know if we can assume these settings are correct for every device using pin 15 to control backlight.

mcrobs:
Hi, I tried this parameter but this has no effect on GPIO15 . GPIO is used as input parameter on ESP32 to select an external storage, therefore I assume it is initialized differently. If you whish I can add this as own config parameter, no deal.

kisvegabor:
If I understood correctly then this pin behavior is specific to ESP32 itself (and not boards). If so, this fix seems correct to me.

@jeremyjh @C47D what do you think?

jeremyjh:
@kisvegabor I looked this up and I agree, my mistake.

mcrobs:
Hi, I would be nice if someone else can verify this. Nevertheless, for this ardutouch case, I need to modify more code, because touchinput and display are using the same spi, but both devices are currently initialized  differently.
disp_spi.c 
`
    ....
    //Initialize the SPI bus
    ret=spi_bus_initialize(HSPI_HOST, &buscfg, 1);
    assert(ret==ESP_OK);
`
versus tp_spi
 `
	//Initialize the SPI bus
	//ret=spi_bus_initialize(HSPI_HOST, &buscfg, 2);
	//assert(ret==ESP_OK);
`

 I think the best is to add a own config session where its possible to select this device, like M5Stack etc.  and to activate in this way all the modifications. What do you mean?

jeremyjh:
@mcrobs  I don't think you need more config for GPIO pin logic as any ESP32 board using that pin should have the same requirement. 

Similarly if touch & display are both using the same SPI pins HSPI should only be initialized once. I would guess it doesn't matter which DMA channel is used (1 or 2 both fine) as long as the init only runs once and the same device is used for both after that.

Still you can certainly set a new device type that has defaults for all the correct pins, that would definitely help the next person down the line.

kisvegabor:
I've merged this improvement. Thank you!

@mcrobs For the other proposed update please open a new pull request or issue to discuss it separately.

PR #51, opened 2020-02-15
New directory structure

C47D:
Necessary changes to separate tft and touch code and also renaming of directories so users doesn't assume this repo only support ili9341 driver.

The configuration of the tft driver and touch driver is now into separated menues:
split_configurations

Display configuration:
display_configuration

Touch configuration:
touch_configuration

I would like to get feedback about this change, maybe @seeul8er, @jeremyjh or anyone else can help with it?


embeddedt:
This doesn't exactly pertain to the directory structure, but since this repository is intended to be used by LittlevGL users, can we move the menu options to the top of the list instead of the bottom?

C47D:
I don't know if it's possible, both TFT and touch menues are user `components` so they go after all the other `official` components.

seeul8er:
Tested it with the ILI9488 & FT6236 and everything works fine. Looks good to me :+1:

PR #53, opened 2020-02-23
correct handling of DMA buffers

jaghatei:
spi_device_queue_trans implicitly allocates capable memory (DMA and 32bit aligned) if provided buffer in spi_transaction_t structure is not.
But only spi_device_get_trans_result sets this free again. So without calling spi_device_get_trans_result there is a memory leak stalling application after just a few calls to spi_device_queue_trans.

### DIFF
--- a/components/lvgl_esp32_drivers/lvgl_tft/disp_spi.c
+++ b/components/lvgl_esp32_drivers/lvgl_tft/disp_spi.c
@@ -95,6 +95,8 @@ void disp_spi_send_data(uint8_t * data, uint16_t length)
     spi_trans_in_progress = true;
     spi_color_sent = false;             //Mark the "lv_flush_ready" NOT needs to be called in "spi_ready"
     spi_device_queue_trans(spi, &t, portMAX_DELAY);
+	spi_transaction_t *ta = &t;
+    spi_device_get_trans_result(spi,&ta, portMAX_DELAY);
 
 }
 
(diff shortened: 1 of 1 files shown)


C47D:
Thanks for the patch, I'm not as familiar with the ESP32 ecosystem as I would like, I'm just gessing here, did you saw the memory leaking when using the idf monitor?

jaghatei:
I found the screen updates stalling after just a few updates. On serial output Watchdog was complaining about task not freeing cpu in time. Higher loglevel ( debug) then revealed that spi_master.c / spi_device_queue_trans / setup_priv_desc was allocating DMA capable memory.
Further analysis showed that memory is freed only by uninstall_priv_desc which is called in spi_device_get_trans_result.

C47D:
Thanks for the explanation, one more question if you don't mind, do you think we should check the return value of `spi_device_get_trans_result`? Just to be sure the transaction is being done OK?

jaghatei:
With respect to the documentation in the source maybe yes:
spi_device_get_trans_result could return ESP_ERR_TIMEOUT and then call shall be retried.
But currently I cannot imagine under what circumstances this would happen and what this means to the application:
Just retry until ESP_OK could end up in infinite loop in app in case bus is blocked.
Just retry x times may not be enough.
Maybe a return != ESP_OK should be reported to higher function and bus and display should be reinitiated to have a defined state.

PR #54, opened 2020-02-23
Allow TFT and Touch to share SPI bus

tonystuart:
This lets you configure the ILI9341 TFT display and the XPT2046 touch controller as devices on a shared SPI bus, freeing the other ESP32 SPI bus for other peripherals.

I didn't see #50 until just now, perhaps any changes that are being considered for that issue would eliminate the need for this pull request.

I tried to make a minimal set of changes that includes:

- Moving the configuration and handling of TP_SPI_CS to the esp-idf SPI layer so that it can be managed as part of the SPI transaction
- Reading the touch X and Y values in a single transaction
- Chaining to an optional user-supplied post_cb in disp_spi.c so that spi_ready can remain a private function
- Splitting the SPI bus and device configuration into a number of smaller functions to provide more options for initializing the bus and devices
- Updating main to initialize and configure the SPI bus if the definition of the MOSI and CLK GPIO pins indicates that they are shared

I think these last two items are the most relevant to #50.

PR #55, opened 2020-02-26
Shared spi bus single commits

tonystuart:
Hi @C47D ,

As discussed in #54 , here are the changes split out as separate commits.

The first commit includes both moving the CS signal management to the SPI layer as well as reading the touch values in a single SPI transaction. I put them both in a single commit because one depends on the other.

The second commit splits out the touch SPI configuration functions into separate functions to allow more configuration options.

The third commit splits out the display SPI configuration functions. It's like the second commit, but for the display.

The fourth commit provides an example of using the changes to share a single SPI bus for both the display and touch controllers.

I tested each of the changes independently with a bread-boarded ESP32 and 2.8 inch touch display with the ILI9341 and XPT2046 on separate SPI buses (e.g. HSPI_HOST and VSPI_HOST).

The final example also works with the controllers on the same SPI bus (e.g. HSPI_HOST), if the MOSI and CLK values in sdkconfig are the same for both devices.

Incidentally, when using a shared SPI bus, it looks like it is necessary to remove the SPI_DEVICE_HALFDUPLEX flag from the display device configuration.

Otherwise the touch handler doesn't receive data from the touch device via MISO (presumably the half duplex flag prevents this).

Removing SPI_DEVICE_HALFDUPLEX also requires reducing the bus speed for the device, but the example still looks great at 20 MHz.


C47D:
Hi,

The `SPI_DEVICE_HALFDUPLEX` flag was used because they werent getting data back from the TFT controller. I am not sure if the effects of setting the SPI bus speed to 20MHz might be more notisable on bigger displays, I will check it with a display I have at home, it's 3.5".

Regards

components/lvgl_esp32_drivers/lvgl_touch/tp_spi.c
```diff
@@ -51,9 +51,9 @@ void tp_spi_init(void)
...
 
 	spi_device_interface_config_t devcfg={
-		.clock_speed_hz=2*1000*1000,           //Clock out at 80 MHz
+		.clock_speed_hz=2*1000*1000,            //Clock out at 2 MHz
 		.mode=0,                                //SPI mode 0
-		.spics_io_num=-1,              //CS pin
+		.spics_io_num=TP_SPI_CS,                //CS pin
 		.queue_size=1,
 		.pre_cb=NULL,
 		.post_cb=NULL,
```
C47D: According to [this](https://esp32.com/viewtopic.php?t=1499) post on the esp32 forum, setting a GPIO number in the `spics_io_num` field limits the number of devices up to three on the spi bus, it suggest to set `spics_io_num` to -1 and handle the cs control on the `pre_cb` and `post_cb` callbacks. Maybe we can try that so the number of slaves on the SPI bus isnt't limited to three. What do you think?

components/lvgl_esp32_drivers/lvgl_touch/xpt2046.c
```diff
@@ -48,10 +48,16 @@ uint8_t avg_last;
...
-    gpio_set_direction(XPT2046_IRQ, GPIO_MODE_INPUT);
-    gpio_set_direction(TP_SPI_CS, GPIO_MODE_OUTPUT);
-    gpio_set_level(TP_SPI_CS, 1);
-
+    gpio_config_t irq_config = {
+        .pin_bit_mask = 1UL << XPT2046_IRQ,
+        .mode = GPIO_MODE_INPUT,
+        .pull_up_en = GPIO_PULLUP_DISABLE,
+        .pull_down_en = GPIO_PULLDOWN_DISABLE,
+        .intr_type = GPIO_INTR_DISABLE,
```
C47D, line 56: Here are you configuring the `XPT2046_IRQ` pin as input only and not as interrupt isn't?

I don't know if lvgl have limitations when reading input devices, maybe in the future we can try configuring the `XPT2046_IRQ` to trigger an interrupt on a touch event.

components/lvgl_esp32_drivers/lvgl_touch/tp_spi.c
tonystuart: That's interesting... unfortunately it looks like you have to modify the esp-idf NO_CS manifest constant in spi_master.c in order to make it work, and if you do that, then it seems like the esp-idf layer will handle more than three devices via the spics_io_num field.

components/lvgl_esp32_drivers/lvgl_touch/xpt2046.c
tonystuart: This one is tricky. Although the pin is called an interrupt signal, it is asserted both when the screen is touched and then later when the value is read. This [post](https://github.com/PaulStoffregen/XPT2046_Touchscreen) gives a pretty good explanation of the situation 

> However, inside your interrupt function, if the display is no longer being touched, any attempt to read the touch position will cause the interrupt pin to create another falling edge. This can lead to an infinite loop of falsely triggered interrupts. Special care is needed to avoid triggering more interrupts on the low signal due to reading the touch position:

> For most applications, regularly reading the touch position from the main program is much simpler.

The library's current approach of polling the interrupt line to see if the screen is being touched and only invoking the SPI transaction if it is seems like a fairly good option, albeit not as CPU efficient as a purely interrupt driven solution.

tonystuart:
I've been meaning to take a look at the esp-idf SPI source code to see why it seems that you can't have two devices on the same bus, one running half-duplex at a higher speed, and one running full-duplex at a lower speed. It seems like it should be possible to do both. Perhaps I don't have it configured quite right.

tonystuart:
Just wanted to mention that I'm going to be travelling through 3/8 and may not have regular internet access.

components/lvgl_esp32_drivers/lvgl_touch/xpt2046.c
C47D: You're right, lvgl reads the input device periodically to know it's state.

C47D:
> I've been meaning to take a look at the esp-idf SPI source code to see why it seems that you can't have two devices on the same bus, one running half-duplex at a higher speed, and one running full-duplex at a lower speed. It seems like it should be possible to do both. Perhaps I don't have it configured quite right.

I want to test this later today.

There are a couple of tft controllers we can add support for next week, and also another touch controller, so this PR will enable us to do so very easily :)

C47D:
I've tested the PR on a Wrover kit v4.1 and also with another ESP32 + ILI9488 in a breadboard, both worked properly, thanks for the help @tonystuart!

PR #75, opened 2020-03-11
created 'user friendly' driver

bzgec:
Idea is to create a 'user friendly' driver meaning that `main.c` would look something like this:
```
...
#include "lvgl/lvgl.h"
#include "lvgl_driver.h"
...
void app_main() {
    lv_init();
    lvgl_driver_init();
...
}
```
I don't know if i configured `CMakeLists.txt` and `component.mk` correctly, but the demo program works (tested on ILI9341 and XPT2046 with shared SPI).
Probably the file/function names are not really the best, but this is only a suggestion...

### DIFF
--- a/components/lvgl_esp32_drivers/lvgl_driver.c
+++ b/components/lvgl_esp32_drivers/lvgl_driver.c
@@ -23,7 +23,9 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
+#ifdef SHARED_SPI_BUS
 static void configure_shared_spi_bus(void);
+#endif  // SHARED_SPI_BUS
 
 /**********************
  *  STATIC VARIABLES
@@ -60,6 +62,7 @@ void lvgl_driver_init(void)
 /**********************
  *   STATIC FUNCTIONS
  **********************/
+#ifdef SHARED_SPI_BUS
 static void configure_shared_spi_bus(void)
 {
 	/* Shared SPI bus configuration */
@@ -87,3 +90,4 @@ static void configure_shared_spi_bus(void)
 	disp_spi_add_device(TFT_SPI_HOST);
 	tp_spi_add_device(TOUCH_SPI_HOST);
 }
+#endif  // SHARED_SPI_BUS
--- a/components/lvgl_esp32_drivers/lvgl_driver.c
+++ b/components/lvgl_esp32_drivers/lvgl_driver.c
@@ -23,6 +23,7 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
+static void configure_shared_spi_bus(void);
 
 /**********************
  *  STATIC VARIABLES
@@ -56,7 +57,10 @@ void lvgl_driver_init(void)
 #endif  // SHARED_SPI_BUS
 }
 
-void configure_shared_spi_bus(void)
+/**********************
+ *   STATIC FUNCTIONS
+ **********************/
+static void configure_shared_spi_bus(void)
 {
 	/* Shared SPI bus configuration */
 	spi_bus_config_t buscfg = {
@@ -83,8 +87,3 @@ void configure_shared_spi_bus(void)
 	disp_spi_add_device(TFT_SPI_HOST);
 	tp_spi_add_device(TOUCH_SPI_HOST);
 }
-
-
-/**********************
- *   STATIC FUNCTIONS
- **********************/
--- a/components/lvgl_esp32_drivers/lvgl_driver.c
+++ b/components/lvgl_esp32_drivers/lvgl_driver.c
@@ -1,0 +1,90 @@
+/**
+ * @file lvgl_driver.c
+ *
+ */
+
+/*********************
+ *      INCLUDES
+ *********************/
+#include "sdkconfig.h"
+#include "lvgl_driver.h"
+
+#include "lvgl_tft/disp_spi.h"
+#include "lvgl_touch/tp_spi.h"
+
+/*********************
+ *      DEFINES
+ *********************/
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+
+/**********************
+ *      MACROS
+ **********************/
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+void lvgl_driver_init(void)
+{
+    /* Interface and driver initialization */
+#ifdef SHARED_SPI_BUS
+	/* Configure one SPI bus for the two devices */
+	configure_shared_spi_bus();
+    
+	/* Configure the drivers */
+	disp_driver_init(false);
+#if CONFIG_LVGL_TOUCH_CONTROLLER != TOUCH_CONTROLLER_NONE
+	touch_driver_init(false);
+#endif
+#else
+	/* Otherwise configure the SPI bus and devices separately inside the drivers*/
+	disp_driver_init(true);
+#if CONFIG_LVGL_TOUCH_CONTROLLER != TOUCH_CONTROLLER_NONE
+	touch_driver_init(true);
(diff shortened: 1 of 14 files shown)


C47D:
Thanks for the PR @bzgec, I did a quick view at it and only have one doubt, there's support for a touch controller that speaks I2C, can this PR consider it as is or we should modify it?

bzgec:
I found 1 typo for the touch controller that speaks I2C (`ft6x06`) in `touch_driver.c`. I don't have that touch controller to test it, but in a few hours I am going to check if it compiles for `ft6x06`.

C47D:
That typo is a tricky one, I also had it once :)

I like how the application main ended up, I only have one suggestion, can we create a new directory named something like `lvgl_driver` in the `lvgl_esp32_drivers` and place the new files in it? (we also will have to update the CMakeList.txt files)

I was also thinking on create a new menuconfig entry (you can see the preview [here](https://forum.littlevgl.com/t/esp32-port-config/1760/4)) and we can add a config to create the demo or not, if we don't create the demo then the project is fully configured so the users can test lvgl with their application more easily, what do you think?

main.c
```c
	esp_register_freertos_tick_hook(lv_tick_task);

#if CONFIG_LVGL_CREATE_DEMO
	demo_create();
#endif
```

components/lvgl_esp32_drivers/lvgl_driver.c
```diff
@@ -0,0 +1,90 @@
...
+#else
+	/* Otherwise configure the SPI bus and devices separately inside the drivers*/
+	disp_driver_init(true);
+#if CONFIG_LVGL_TOUCH_CONTROLLER != TOUCH_CONTROLLER_NONE
+	touch_driver_init(true);
+#endif
+#endif  // SHARED_SPI_BUS
+}
+
+void configure_shared_spi_bus(void)
```
C47D: Can we make this function static?

components/lvgl_esp32_drivers/lvgl_spi_conf.h
```diff
@@ -0,0 +1,72 @@
...
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+/*********************
+ *      DEFINES
+ *********************/
+// DISPLAY PINS
+#define DISP_SPI_MOSI CONFIG_LVGL_DISP_SPI_MOSI
```
C47D: This is already defined in `disp_spi.h`, can we use those?

components/lvgl_esp32_drivers/lvgl_spi_conf.h
```diff
@@ -0,0 +1,72 @@
...
+#define DISP_SPI_CLK CONFIG_LVGL_DISP_SPI_CLK
+#define DISP_SPI_CS CONFIG_LVGL_DISP_SPI_CS
```
C47D, line 23: This symbols are already defined in `lvgl_tft/disp_spi.h`, can we use those?

components/lvgl_esp32_drivers/lvgl_spi_conf.h
```diff
@@ -0,0 +1,72 @@
...
+
+// TOUCHPAD PINS
+#define TP_SPI_MOSI CONFIG_LVGL_TOUCH_SPI_MOSI
+#define TP_SPI_MISO CONFIG_LVGL_TOUCH_SPI_MISO
+#define TP_SPI_CLK  CONFIG_LVGL_TOUCH_SPI_CLK
+#define TP_SPI_CS   CONFIG_LVGL_TOUCH_SPI_CS
```
C47D, line 29: This symbols are already defined in `lvgl_touch/touch_driver.h`, can we use those?

bzgec:
> can we create a new directory named something like `lvgl_driver` in the `lvgl_esp32_drivers` and place the new files in it?

I kinda like the way it is. Because it is in root of `lvgl_esp32_drivers` folder, therefore in includes both display driver and touch driver. Does this make sense?

components/lvgl_esp32_drivers/lvgl_driver.c
bzgec: Sure, it is done.

components/lvgl_esp32_drivers/lvgl_spi_conf.h
bzgec: I thought i removed those from`lvgl_tft/disp_spi.h`. I like that all pins are defined in one file (for instance if you don't want to use `menuconfig` any more).

components/lvgl_esp32_drivers/lvgl_spi_conf.h
bzgec: I thought i removed those from`lvgl_touch/tp_spi.h`. I like that all pins are defined in one file (for instance if you don't want to use `menuconfig` any more).

bzgec:
> I was also thinking on create a new menuconfig entry (you can see the preview here) and we can add a config to create the demo or not, if we don't create the demo then the project is fully configured so the users can test lvgl with their application more easily, what do you think?

I think that this is not necessary, because when user doesn't want to use demo he/she can just comment out the function call (`demo_create();`) and also the code is more clear/simple/transparent (sorry couldn't fine the correct expression) when there are less `#if ...` and `#endif`.

C47D:
> I kinda like the way it is. Because it is in root of lvgl_esp32_drivers folder, therefore in includes both display driver and touch driver. Does this make sense?

Agree

> I think that this is not necessary, because when user doesn't want to use demo he/she can just comment out the function call (demo_create();) and also the code is more clear/simple/transparent (sorry couldn't fine the correct expression) when there are less #if ... and #endif.

True, let me test it with the ILI9488 and ST7789 display controllers i have and i will merge it.

C47D:
I've tested it with ILI9488 display (no touch) and it works but the ST7789 display shows the colors inverted....

components/lvgl_esp32_drivers/lvgl_driver.c
C47D: I get the `unused function` warning when no using a touch controller, do you think we can do something like (or define `SHARED_SPI_BUS` and do `#if SHARED_SPI_BUS`):
```c
#if CONFIG_LVGL_TOUCH_CONTROLLER != TOUCH_CONTROLLER_NONE
static void configure_shared_spi_bus(void);
#endif

...

#if CONFIG_LVGL_TOUCH_CONTROLLER != TOUCH_CONTROLLER_NONE
static void configure_shared_spi_bus(void)
{
    /* implementation */
}
#endif
```

What do you think?

components/lvgl_esp32_drivers/lvgl_spi_conf.h
```diff
@@ -0,0 +1,72 @@
...
+#if CONFIG_LVGL_TOUCH_CONTROLLER_SPI_HSPI == 1
+#define TOUCH_SPI_HOST HSPI_HOST
+#else
+#define TOUCH_SPI_HOST VSPI_HOST
+#endif /*CONFIG_LVGL_TOUCH_CONTROLLER_SPI_HSPI == 1*/
+
+// Detect the use of a shared SPI Bus and verify the user specified the same SPI bus for both touch and tft
+#if (CONFIG_LVGL_TOUCH_CONTROLLER == 1 || CONFIG_LVGL_TOUCH_CONTROLLER == 3) && TP_SPI_MOSI == DISP_SPI_MOSI && TP_SPI_CLK == DISP_SPI_CLK
+#if TFT_SPI_HOST != TOUCH_SPI_HOST
+#error You must specifiy the same SPI host for both display and input driver
```
C47D: Typo in speifiy, should be specify

bzgec:
I fixed typo and put `configure_shared_spi_bus()` in `#ifdef SHARED_SPI_BUS` statement. 

> thanks for the work you been doing

My pleasure

C47D:
Let's merge it and fix the ST7789 in another PR, now I'm wondering if the display had always displayed inverted colors and I didn't noticed it before.

Issue #79, opened 2020-03-16
Readme.md should be enhanced regarding adjustments in Makefile


an-erd:
The Readme says in section [Readme.md#makefile](https://github.com/littlevgl/lv_port_esp32#makefile) to add the following line(s):

> EXTRA_COMPONENT_DIRS := externals/lv_port_esp32/components

If only this directory is added, the Kconfig files in the following files are not reflected in `menuconfig`:
> ./externals/lv_port_esp32/components/lvgl_esp32_drivers/lvgl_tft/Kconfig
> ./externals/lv_port_esp32/components/lvgl_esp32_drivers/lvgl_touch/Kconfig

and thus compilation does not work.

Adding the following line works for me:

> EXTRA_COMPONENT_DIRS := externals/lv_port_esp32/components externals/lv_port_esp32/components/lvgl_esp32_drivers externals/lv_port_esp32/components/lvgl_esp32_drivers/lvgl_touch externals/lv_port_esp32/components/lvgl_esp32_drivers/lvgl_tft

C47D:
Hi @an-erd,

You're right, the README haven't been totally updated as should be. As user what do you find the most easy way to use this project as a template for your personal project:
- By doing what you did based on the instructions of https://github.com/littlevgl/lv_port_esp32#install-this-project-as-a-library-submodule-in-your-own-project
- Deleting the demo_create function call and add your application like any other esp32 idf based project?

I ask because #77 have the same problem

an-erd:
Hi, sorry I missed the point in #77. Please see my comment in #81,

an-erd:
I'm using ESP-IDF with LittlevGL as a submodule, so your first option. I was asking why you put this in externals and not components? In components I've all other submodules, so if there's not a reason for that I'd propose to move to submodules.

So far with the above adjustment in the  makefile it's working pretty good, integration was very easy and straightforward. I currently only have a M5Stack with an ILI9341 without a touchpad. I'm using this code called from main.c I took from the examples, and doing the work in  `lv_testdisplay_create();` :

>void initialize_lv()
>{
>    lv_init();
>    lvgl_driver_init();
>    static lv_color_t buf1[DISP_BUF_SIZE];
>    static lv_color_t buf2[DISP_BUF_SIZE];
>    static lv_disp_buf_t disp_buf;
>    lv_disp_buf_init(&disp_buf, buf1, buf2, DISP_BUF_SIZE);
>    lv_disp_drv_t disp_drv;
[...]

PR #81, opened 2020-03-16
Update README instructions

C47D:
Based in #79

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,10 +82,12 @@ idf_component_register(
 If you are using make, you only need to add the EXTRA_COMPONENT_DIRS in the root Makefile of your project:
 ```Makefile
 PROJECT_NAME := blink
-EXTRA_COMPONENT_DIRS := components/lv_port_esp32/components              \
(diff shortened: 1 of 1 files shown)


an-erd:
In PR #79 the EXTRA_COMPONENT_DIRS needs to be a single line or split using "\\"

C47D:
@an-erd did you meant like that?

an-erd:
Yes, that works for the makefile.

C47D:
Hi @an-erd, 

I've just updated the README, so now instead of creating a directory named `externals` now it's named `components`. I've tested both CMake and Makefiles and had no success.

With the CMake approach I had the following error:
image

With the Makefile approch the menuconfig appeared but I didn't find the new menues were they should be.

Can you test the Makefile approach on your side to check if it works, I'm sure I'm doing something wrong...

Regards
Carlos

C47D:
@an-erd Nevermind, i found the issues and now everything seems to be working as expected :)

Issue #39, opened 2019-12-07
Program crashes when adding wifi support 


cbarransunny:
We use GitHub issues for development related discussions.
Please use the [forum](https://forum.littlevgl.com/) to ask questions.

**Describe the issue**
added wifi station code from esp-idf mqtt example:
https://github.com/espressif/esp-idf/blob/release/v3.3/examples/protocols/mqtt/tcp/main/app_main.c#L171

**Code to reproduce the issue**

```c
  esp_register_freertos_tick_hook(lv_tick_task);
  xTaskCreate(display_task, "display_task", 2048, NULL, 5, NULL); 
  esp_err_t error_test = nvs_flash_init();
  if(error_test == ESP_OK)
  {
    wifi_init_sta();
  }
  while(1)
  {
		vTaskDelay(1000/portTICK_PERIOD_MS);
	}
}

static void display_task(void *pvParameter)
{
  Demo();
  while(1) {
		vTaskDelay(5/portTICK_PERIOD_MS);
		lv_task_handler();
	}

}
```
**Expected Results**
The display shows to demo and wifi connects
**Actual Results**
Guru Meditation Error: Core 0 panic'ed (Cache disabled but cached memory region accessed)
**ESP32 Chip version**

**ESP-IDF version**
3.3
**Development kit used**
NodeMCU 32s
**Development machine OS**
Windows 10, Platoformio
**Compilation warnings/errors (if available)**

**If possible, copy the compilation log into a file and attach it here**

embeddedt:
Not stale. (I can't help with this issue as I don't use ESP32.)

embeddedt:
@kisvegabor @amirgon Perhaps one of you can help with this?

amirgon:
Where is the stack trace?
On a crash report, the very minimum is providing a parsed stack trace.

justbarran:
Problem was solved by running Littlevgl only on core 1 while wifi and everything else runs on core 0. And using esp_timer_start_periodic(_lv_tick_timer, LV_TICK_MS*1000) to run the ticker.

embeddedt:
It's probably a concurrency issue (LittlevGL needs you to [wrap calls with some type of synchronization primitive](https://docs.littlevgl.com/en/html/porting/os.html) to work with multithreaded systems).

jeremyjh:
[...]

PR #84, opened 2020-03-23
More complete example working with different peripherals

nicofossa:
The example has been rewrote to use tasks.
I spent almost a night trying to understand a memory corruption, but this seems to work pretty well.
Tested successfully with high memory and CPU usage, high network traffic and concurrent i2s communication.
It pins the `guiTask` to only one core in order to prevent problem with SPI communication.
Uses `esp_timer_start_periodic` to call `lv_tick_task` periodically.
Uses Semaphores to show how to synchronize with different other tasks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,7 @@ idf_component_register(SRCS ${SOURCES}
 
 target_compile_definitions(${COMPONENT_LIB PRIVATE LV_CONF_INCLUDE_SIMPLE=1)
 ```
+Please, note that if your project require the use of the `nvs_flash` module, it should be put in the `REQUIRES` list.
 
 #### Makefile
 If you are using make, you only need to add the EXTRA_COMPONENT_DIRS in the root Makefile of your project:
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ idf_component_register(SRCS ${SOURCES}
 
 target_compile_definitions(${COMPONENT_LIB PRIVATE LV_CONF_INCLUDE_SIMPLE=1)
 ```
-Please, note that if your project require the use of the `nvs_flash` module, it should be put in the `REQUIRES` list.
+Please note that if your project require the use of the `nvs_flash` module \(for example required by WiFi\), it should be put in the `REQUIRES` list.
 
 #### Makefile
 If you are using make, you only need to add the EXTRA_COMPONENT_DIRS in the root Makefile of your project:
--- a/main/main.c
+++ b/main/main.c
@@ -45,11 +45,11 @@ void app_main() {
     
     //If you want to use a task to create the graphic, you NEED to create a Pinned task
     //Otherwise there can be problem such as memory corruption and so on
-	xTaskCreatePinnedToCore(guiTask, "gui", 4096*2, NULL, 0, NULL, 1);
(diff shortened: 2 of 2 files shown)


C47D:
Hi @nicofossa,

Thanks for the PR, I'm pretty new to the esp32 special features like pinning a task to a dedicated CPU, but it looks very interesting.

What hardware are you using to test it? I got a couple of boards to test this on.

Regards

nicofossa:
Hi!
Esp32 has 2 CPU that can run concurrently, so the "guiTask" may run on the two core without any restriction. It seems that if we access the SPI peripheral a first time with one CPU and a second time with the other one, a strange memory corruption happens, maybe related to DMA stuff.
I didn't deepen in this error much, but in issue #39 was proposed to pin the task, and it actually worked.
As different task may run concurrently on different CPUs and access lvgl stuff, the Semaphore is needed to synchronize them.
I tested this code on a ESP32-WROOM-32, on ESP32S development board, with external ILI9341.
I also run the "stress test" example included, and didn't run into any issue.

nicofossa:
I also think that with this, #84 can also be closed.

C47D:
Hi,

> Esp32 has 2 CPU that can run concurrently, so the "guiTask" may run on the two core without any restriction. It seems that if we access the SPI peripheral a first time with one CPU and a second time with the other one, a strange memory corruption happens, maybe related to DMA stuff.
> I didn't deepen in this error much, but in issue #39 was proposed to pin the task, and it actually worked.

Thanks for the explanation, I just saw that the lvgl thread is pinned to the applications core meanwhile the wifi can run on the other core.

> As different task may run concurrently on different CPUs and access lvgl stuff, the Semaphore is needed to synchronize them.

Afaik all the lvgl related code must run on a single thread, I will need to investigate this further.

> I tested this code on a ESP32-WROOM-32, on ESP32S development board, with external ILI9341.

I will test it with the ESP Wrover Kit 4.1 tonight, then I can merge it.

> I also think that with this, #84 can also be closed.

Yeah, I was planning to add an example with WiFi on it and use lvgl to show some information, but I haven't had time to do so.

Regards

embeddedt:
> Afaik all the lvgl related code must run on a single thread, I will need to investigate this further.

LittlevGL can run on multiple threads, provided that a [mutex](https://docs.littlevgl.com/en/html/porting/os.html#tasks-and-threads) is used to guard API calls. Typically a dedicated thread runs `lv_task_handler`, and other threads can then work with objects.

nicofossa:
> LittlevGL can run on multiple threads, provided that a [mutex](https://docs.littlevgl.com/en/html/porting/os.html#tasks-and-threads) is used to guard API calls. Typically a dedicated thread runs `lv_task_handler`, and other threads can then work with objects.

While this is true, it seems that calling `lv_task_handler` "sequentially" (not concurrently) from two different CPUs crashes the ESP32, corrupting the memory.

C47D:
I got the following warning:
```
main.c:87:25: warning: initialization of 'void (*)(void *)' fro incompatible pointer type 'void (*)(void)' [-Wincompatible-pointer-types]
.callback = &lv_tick_task,
```

PR #87, opened 2020-03-24
Change default backlight GPIO

seeul8er:
For ER-TFTM035 and the default one to GPIO27.
Prev. GPIO21 is the I²C interface used by some touch controllers. 
GPIO27 has no important dedicated functionality


C47D:
Thanks  @seeul8er

PR #94, opened 2020-04-13
Added SSD1306 and SH1107 display controller, added ILI9341, SH1107 and SSD1306 sleep mode

an-erd:
- Added display controller SSD1306 
- Added display controller SH1107
- Added predefined device (Wemos Lolin ESP32 OLED 128x64 with SSD1306)
- Added predefined device (M5Stick with SH1107)
- Merged code for display orientation (commits 900c1d1 and 42c2fd2)
- Added SSD1306, ILI9341, SH1107 sleep mode
- Added SSD1306 and SH1107 inverse display, and display rotation (portrait/landscape)
- Kconfig cleanup
- Updated README.md


C47D:
Thanks a lot for the contibution @an-erd, I tested your fork over the weekend with the SSD1306 controller and it works great.

an-erd:
Yes, that's a good idea. I will implement it that way.

I use the set_px_cb because one pixel is just one bit on the monochrome displays and so it's only used for these monochrome displays. And with this the buffer can be send to the display controller directly and thus efficiently. The rounder is necessary because both display driver use columns and _pages_ (which is a height  of 8 px und thus one byte), and thus we need to send complete pages to the display controller.

C47D:
> I use the set_px_cb because one pixel is just one bit on the monochrome displays and so it's only used for these monochrome displays. And with this the buffer can be send to the display controller directly and thus efficiently. The rounder is necessary because both display driver use columns and pages (which is a height of 8 px und thus one byte), and thus we need to send complete pages to the display controller.

Thanks for the explanation, I'm a bit behind the internal working of lvgl, need to do some toy projects to get a lot more familiar with it :)

Regards

PR #99, opened 2020-04-15
Move LVGL configuration (Fonts, memory, themes) to Kconfig

an-erd:
Enable configuration from lv_conf.h to Kconfig for:
- Fonts
- Themes
- Memory


an-erd:
Comments
- sdkconfig now should be removed from github now
- change from #98 is included here (is not visible in master, yet)

C47D:
Seems all correct, we can fix any issues later, thanks for the continous work @an-erd.

PR #102, opened 2020-04-25
Fixed default display height for ER-TFT, Changed default CS pin

seeul8er:
Prev. CS was equal to CLK


C47D:
Thanks for the fix @seeul8er.

PR #109, opened 2020-05-02
ILI9488: Fix low DMA crash - retry on allocation failure

seeul8er:
In case of high DMA usage, the driver may not be able to allocate enough DMA capable memory resulting in a crash. Retry loop till enough memory becomes available.

Example of high DMA usage: High rate of HTTP requests & responses (>2048 bytes) via WiFi while updating big portions of the screen.

### DIFF
--- a/components/lvgl_esp32_drivers/lvgl_tft/ili9488.c
+++ b/components/lvgl_esp32_drivers/lvgl_tft/ili9488.c
@@ -121,10 +121,11 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
     uint32_t size = lv_area_get_width(area) * lv_area_get_height(area);
 
     lv_color16_t *buffer_16bit = (lv_color16_t *) color_map;
-    uint8_t *mybuf = (uint8_t *) heap_caps_malloc(3 * size * sizeof(uint8_t), MALLOC_CAP_DMA);
-    while (mybuf == NULL) {
+    uint8_t *mybuf;
+    do {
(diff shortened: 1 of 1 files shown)


components/lvgl_esp32_drivers/lvgl_tft/ili9488.c
```diff
@@ -122,6 +122,9 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
 
     lv_color16_t *buffer_16bit = (lv_color16_t *) color_map;
     uint8_t *mybuf = (uint8_t *) heap_caps_malloc(3 * size * sizeof(uint8_t), MALLOC_CAP_DMA);
+    while (mybuf == NULL) {
+        mybuf = (uint8_t *) heap_caps_malloc(3 * size * sizeof(uint8_t), MALLOC_CAP_DMA);
```
embeddedt, line 126: How about a `do { } while(mybuf == NULL)` loop to avoid repeating the same line twice?

seeul8er: I agree :+1: I will fix it. Thought about adding some kind of warning message as well. Without it the UI would just freeze without showing any kind of hint why.
So maybe change it to something like: 
`do { ... if (mybuf == NULL) {ESP_LOGW()} } while(mybuf == NULL)`. Still not beautiful either, but at least some debugging info.

C47D:
Looks good to me, I think this is the only driver that allocates memory hehe.

PR #110, opened 2020-05-04
Dev 7.0

C47D:
So we can track the commits being added to the `dev-7.0` branch. Wasn't able to add @barbiani as reviewer, but you're welcome to let us know when this is good to go.

There are some other tasks I would like to work on:

- [x] Indev, I haven't tested any touch drivers.
- [x] Multiple devices (touch and display controllers) on the same bus, I2C should be easier than SPI.
- [x] Cleanup and comment the code on `lvgl_esp32_drivers` directory, so is easier for users to add support for their needs.
- [x] Whatever fix we found on the SPI data transfers.
- [x] Some displays doesn't need CS control (some ST7789 boards doesn't have it available)


components/lvgl/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 file(GLOB_RECURSE SOURCES lvgl/src/*.c)
 idf_component_register(SRCS ${SOURCES}
-                       INCLUDE_DIRS . lvgl)
+                       INCLUDE_DIRS . lvgl lvgl/src/lv_themes)
```
kisvegabor: Why `lv_themes` is special?

components/lvgl/Kconfig
```diff
@@ -25,22 +29,25 @@ menu "LittlevGL (LVGL) configuration"
...
+		config LVGL_DEFAULT_FONT_MONTSERRAT12
+			select LVGL_FONT_MONTSERRAT12SUBPIX
+			bool "Montserrat 12 sub-pixel"
+		config LVGL_DEFAULT_FONT_MONTSERRAT16
+			select LVGL_FONT_MONTSERRAT16
+			bool "Montserrat 16"
+		config LVGL_DEFAULT_FONT_MONTSERRAT22
+			select LVGL_FONT_MONTSERRAT22
+			bool "Montserrat 22"
+		config LVGL_DEFAULT_FONT_MONTSERRAT28
```
kisvegabor: We have fonts up to 48 px.  Is there any reason why the rest was skipped or this part is just not finished yet?

C47D:
> Thanks, looks good in general. Although I also can't tell the correctness of the ESP related parts.
> I found only 2 minor things.

Sure, the config files need to be updated, also the README.

What things did you found?

components/lvgl/CMakeLists.txt
C47D: When testing with monochrome displays the monochrome theme wasn't found, I think I can solve it bu using `#include "lvgl/src/lv_themes"` in the main file?

components/lvgl/Kconfig
C47D: It is not finished yet.

C47D:
@kisvegabor Is it possible for LVGL to be 'aware' of any display orientation change? I mean, if the application starts with the display being horizontal and then the application needs to change it to vertical, can LVGL display correctly in the new setup?

I ask this because I'm updating the display orientation code, but it's set at the initialization and not possible to change at run time.

components/lvgl/CMakeLists.txt
C47D: Fixed with b88fcbd

C47D:
Sure, we can add an option for that. Any name suggestions?

davidjade:
I may have a few more suggestions but some of them might be more radical due to what I am working on. 

I have a working driver for the FT81x EVE2 from FTDI*. These are coprocessor displays that have GPU capabilities but I wanted to use lvgl widgets on them, so I wrote a driver which treats it like a big memory mapped display. These displays can use a number of bitmap formats natively so it would be possible to drive them as RGB565, RGB323, or even grayscale or monochrome.

I currently have a 800x480 display working in RGB565 but the refresh rate is only about 3 fps (30Mhz SPI is the max but I have pushed it to 32 sucessfully). I am working toward using QuadSPI though to speed that up and expect to get it to around 10 fps, which of course requires quite a bit more change to disp_spi. One issue I have faced is that the FT81x SPI protocol requires a lot of read/write transactions, so I've had to add read capability to disp_spi too. And bus locking to speed up multiple transactions, etc... I'm not sure that many others will want this though as most people likely pick the FT81x to use the coprocessor to do the work but it's native widget capabilities are quite a bit less capable than lvgl.

One interesting aspect of my driver is that it would be possible to just give some of the screen over to lvgl and also still use the native FT81x coprocessor capabilities too. Think of it as a "window" or island of lvgl. This would make it possible for instance to use a lvgl Chart object alongside native FT81x widgets, etc... I can map one or more rectangles of screen as lvgl controlled regions.

The FT81x also has a touch screen but I haven't started working on that yet. It's not a separate chipset but integrated into the FT81x protocol so it will be one SPI driver that does double-duty.

davidjade:
Here is another thought. Right now there is disp_spi_send_data() and disp_spi_send_color() but they are nearly identical except for queuing vs. polling and handling of spi_color_sent status. I think they should be merged into one function and the various behaviors controlled by flags. I might also need additional control over some things.

For instance, one of the issues that I face with my new display is that CS has to be held low over multiple transactions - its part of the protocol for this display. Unless I want to merge/copy cmds + color data into one large duplicate buffer I have to use multiple transactions and control the CS line myself. This leads to me also wanting to control the spi_color_sent behavior myself too since I only want to assert it when everything is done, along with possibly acquiring the SPI bus too until all transactions have completed to speed things up. So having more granular ways to control the SPI behaviors would be nice for me.

Then there is the issue with retrieving transaction results and freeing DMA buffer I logged earlier: #112. Besides the need to free potential DMA allocations, I might also need to be able to tell when all results have come back - not just when all transaction have been sent.

Again, not sure this makes sense in the bigger picture with typically simpler displays, but I am finding that I might need more significant changes/re-write to disp_spi to suit the FT81x. I'm working on doing those things for myself but I am still working off the lvgl 6 base at this point.

C47D:
Hi @davidjade,

Thanks for the suggestions.

> I currently have a 800x480 display working in RGB565 but the refresh rate is only about 3 fps (30Mhz SPI is the max but I have pushed it to 32 sucessfully). I am working toward using QuadSPI though to speed that up and expect to get it to around 10 fps, which of course requires quite a bit more change to disp_spi. One issue I have faced is that the FT81x SPI protocol requires a lot of read/write transactions, so I've had to add read capability to disp_spi too. And bus locking to speed up multiple transactions, etc... I'm not sure that many others will want this though as most people likely pick the FT81x to use the coprocessor to do the work but it's native widget capabilities are quite a bit less capable than lvgl.

I'm actually working with `disp_spi`, `touch_spi`, (`disp_i2c` and `touch_i2c`) as well, trying to make them lighter by removing duplicated code. Both `disp_spi` and `touch_spi` have functions to initialize the SPI bus, this was also done on `lvgl_driver` when configuring a shared SPI bus.

Now the functions `bool lvgl_spi_driver_init(int host, int miso_pin, int mosi_pin, int sclk_pin, int max_transfer_sz);` and `bool lvgl_i2c_driver_init(int port, int sda_pin, int scl_pin, int speed);` (they may need a better naming) are the only ones to init the SPI and I2C busses, they are defined and used in `lvgl_driver.c`.

We can improve the spi_device related functions by adding the flags you requested here:
> Would it be better to have a flag to control this on a per-transaction basis?

> For instance, one of the issues that I face with my new display is that CS has to be held low over multiple transactions - its part of the protocol for this display. Unless I want to merge/copy cmds + color data into one large duplicate buffer I have to use multiple transactions and control the CS line myself. This leads to me also wanting to control the spi_color_sent behavior myself too since I only want to assert it when everything is done, along with possibly acquiring the SPI bus too until all transactions have completed to speed things up. So having more granular ways to control the SPI behaviors would be nice for me.

For this I think we can handle the nCS line by ourselfs by using the `pre_cb` and `post_cb` transaction callbacks, we can define a default set of callbacks but also let the user overwrite them with custom set of callbacks, something like this.

```c
void lvgl_spi_register_custom_pre_cb(transaction_cb_t user_pre_cb);
void lvgl_spi_register_custom_post_cb(transaction_cb_t user_post_cb);
```

> Here is another thought. Right now there is disp_spi_send_data() and disp_spi_send_color() but they are nearly identical except for queuing vs. polling and handling of spi_color_sent status. I think they should be merged into one function and the various behaviors controlled by flags. I might also need additional control over some things.

For this we could add a flag argument and also a void* argument so you can pass a context to the functions. I don't know if you also would need to 'overwrite' the default behaviour.

> Again, not sure this makes sense in the bigger picture with typically simpler displays, but I am finding that I might need more significant changes/re-write to disp_spi to suit the FT81x. I'm working on doing those things for myself but I am still working off the lvgl 6 base at this point.

Is it possible for you to port your work to LVGL v7 by improving this repo and try to add as little patches to your own fork to handle your special display?

What is certain is that we need to improve the code handling transactions, I will remove all the unused code tonight, I'm at the office right now and didn't bring my dev kits :P.

davidjade:
Rather than duplicate my code example here, I'll just point to #112 (where I put a code example) and add additional comments here.

> For this I think we can handle the nCS line by ourselfs by using the pre_cb and post_cb transaction callbacks, we can define a default set of callbacks but also let the user overwrite them with custom set of callbacks...

The issue I have is that CS need to be held over multiple transactions so pre_cb/post_cb might work if I keep track of everything but really controlling it it something that happens over a sequence of calls to disp_spi, at the next level higher (within my flush function currently). I'm not sure yet what the best approach is and if I can eliminate that need, I will but I think this is going to cause even more changes to the SPI transaction handling (i.e. using spi_transaction_ext_t along with CMD and ADDR fields)

For right now my plan is to continue to hack away but I'm not opposed to contributing things back. I think it makes sense to keep in touch to see what parts can be shared. For instance, I think my example code in #112 would work right now for all the other drivers. I also think that my upcoming plans/changes to handle CS would also work fine with other drivers and not even require anyone else to change - that's my goal anyway. I think my approach to adding a disp_spi_send_data_ex that has more options and that the existing functions delegate too is a nice approach because it allows for any driver to take over more of the decisions if they need to while maintaining compatibility for existing drivers.

C47D:
I haven't been able to update code in this branch. @barbiani, can you try this branch? I've updated the code that you changed on your fork.

C47D:
@davidjade Your improvements to the SPI transfers have been included in this PR, can you test them? The demo code is part of the new demos in lv_example, I would suggest to clone this PR just to check the touch and display controllers are working for you.

davidjade:
A few things I noticed:

From running menuconfig:
1. When selecting the FT81x, the resolution is not set. I'm not positive but almost all FT81x displays are 800x480 so maybe that should be preloaded when it is selected?
2. When selecting the FT81x the MISO pin is not enabled. It should be or it won't work.
3. a GPIO for DC is still selected when using the FT81x - it is not used. Same for reset pin. Should these be hidden for the FT81x?
4. Selecting the VSPI SPI bus did not then set the standard ESP32 pins for VSPI (5, 18, 19, 23)
5. I think I saw recent work to hide orientation for displays that don't support it - this is the case for the FT81x, at least for now.

I did not change any other LvGL menuconfig options before building other than enabling the FT18X display and touch.

When building I get:
../components/lvgl_esp32_drivers/lvgl_driver.h:41:2: error: #error "No display controller selected"

Looking in there, it seems the code for the FT81x never made it in that file. Note that the quick fix of defining DISP_BUF_SIZE for the FT81x there is not quite right, since it is defined elsewhere and it currently needs to be that way for other files and symbol visibility.

Hacking around though I got past this only to then get more build errors, like "No protocol defined for touch controller". Not sure what is up here but remember that the FT81x does not have a separate touch controller. I see the code that tries to init either an SPI or I2C controller but there needs to be a case to handle no separate touch controller but with touch still enabled.

So I put in a few hacks to get past these issues and it now builds. After this I was able to flash the new demo.

However, things were not working and the new logging showed that the MISO pin wasn't being set. I found that that CONFIG_LVGL_DISPLAY_USE_SPI_MISO that is in lvg_spi_conf.h had a typo (as CONFIG_LV_DISPLAY_USE_SPI_MISO), which if course causes the MISO pin to not be set. 

After this I got it to run but I get quite a lot of screen corruption - not sure why but I suspect it's maybe just SPI corruption as lowering the clock rate makes it go away. My breadboard wiring has been pretty stable with the previous demos, etc... but for this version I had to back off the clock rate quite a lot to get it to work. So much so, it's now too slow to really use. 

It seems to be so sensitive now though that the FT81X setup SPI commands are often corrupted, which stops the display from working at all. It's hit or miss whether any given power cycle makes it work. This is so strange to me as the v6 version is pretty rock-solid and it doesn't make sense to me since this is just start up code that should not have changed. I will try and scope the two version to spot any changed if I can.

Once it does work though there's still the odd corruption. If I had to guess it's that maybe LvGL v7 is doing a lot more smaller transfers in the new version than v6 and my breadboard wiring is just showing more instances of SPI corruption but I'm really not sure. I have not looked closely at the merge to know if anything else changed with the core of the SPI and FT18X code from what I first wrote, so I'm not sure what other than my wiring is maybe more sensitive to whatever v7 is doing. I will take a closer look though - maybe there is another SPI change somewhere in the merge that is having an unintended effect.

So it works, sort of, with a bit of hacking to correct a few problems but even then it's not nearly as stable as v6 and I don't really know why at this point. Some of this is likely breadboard wiring but it sure does seem way more sensitive to that now as the best I could run it was 1/2 the clock speed compared to v6 demos.

C47D:
Hi @davidjade,

Thanks for taking the time to test, fix and let me know the issues you had, it's very appreciated. I will answer between lines.

> From running menuconfig:
> 1. When selecting the FT81x, the resolution is not set. I'm not positive but almost all FT81x displays are 800x480 so maybe that should be preloaded when it is selected?

> 2. When selecting the FT81x the MISO pin is not enabled. It should be or it won't work.

This should be fixed with b8fda08, but I checked the lvgl_touch/Kconfig file and there is not options to set the SPI pins when selecting the FT81x controller, below you mention that the same interface is used for display and touch, so let me know what approach works best for you:
- Add options to set the touch Kconfig
- Detect the FT81X controller as special case in the `lvgl_driver_init` and configure the SPI with all the pins configured in the display Kconfig.
- Any other you suggest

> 3. a GPIO for DC is still selected when using the FT81x - it is not used. Same for reset pin. Should these be hidden for the FT81x?

> 4. Selecting the VSPI SPI bus did not then set the standard ESP32 pins for VSPI (5, 18, 19, 23)

This option is not meant to choose the pins for VSPI, we need to know which of the two spi bus to initialize.

> 5. I think I saw recent work to hide orientation for displays that don't support it - this is the case for the FT81x, at least for now. What I mean is, maybe it should be hidden for the FT81X too.

I will take a look at it, I thought it was only enabled for ILI9xxx controllers and two others (monochrome)

> I did not change any other LvGL menuconfig options before building other than enabling the FT18X display and touch.
> 
> When building I get:
> ../components/lvgl_esp32_drivers/lvgl_driver.h:41:2: error: #error "No display controller selected"
> 
> Looking in there, it seems the code for the FT81x never made it in that file. Note that the quick fix of defining DISP_BUF_SIZE for the FT81x there is not quite right, since it is defined elsewhere and it currently needs to be that way for other files and symbol visibility.

Let me know if 7307ad0 fixes it.

> Hacking around though I got past this only to then get more build errors, like "No protocol defined for touch controller". Not sure what is up here but remember that the FT81x does not have a separate touch controller. I see the code that tries to init either an SPI or I2C controller but there needs to be a case to handle no separate touch controller but with touch still enabled.

For now db0d1b2 selects SPI when you choose the FT81x touch controller. See my suggestions in 2.

> So I put in a few hacks to get past these issues and it now builds. After this I was able to flash the new demo.
> 
> However, things were not working and the new logging showed that the MISO pin wasn't being set. I found that that CONFIG_LVGL_DISPLAY_USE_SPI_MISO that is in lvg_spi_conf.h had a typo (as CONFIG_LV_DISPLAY_USE_SPI_MISO), which if course causes the MISO pin to not be set.

Typo fixed in b8fda08

> After this I got it to run but I get quite a lot of screen corruption - not sure why but I suspect it's maybe just SPI corruption as lowering the clock rate makes it go away. My breadboard wiring has been pretty stable with the previous demos, etc... but for this version I had to back off the clock rate quite a lot to get it to work. So much so, it's now too slow to really use.
> 
> It seems to be so sensitive now though that the FT81X setup SPI commands are often corrupted, which stops the display from working at all. It's hit or miss whether any given power cycle makes it work. This is so strange to me as the v6 version is pretty rock-solid and it doesn't make sense to me since this is just start up code that should not have changed. I will try and scope the two version to spot any changed if I can.

I had issues when using breadboard and long wiring when trying to run the ILI9488 display controller at 80MHz, which is overclocked, as you said the startup configuration haven't changed so it should work as with LVGL v6.

> Once it does work though there's still the odd corruption. If I had to guess it's that maybe LvGL v7 is doing a lot more smaller transfers in the new version than v6 and my breadboard wiring is just showing more instances of SPI corruption but I'm really not sure. I have not looked closely at the merge to know if anything else changed with the core of the SPI and FT18X code from what I first wrote, so I'm not sure what other than my wiring is maybe more sensitive to whatever v7 is doing. I will take a closer look though - maybe there is another SPI change somewhere in the merge that is having an unintended effect.
> 
> So it works, sort of, with a bit of hacking to correct a few problems but even then it's not nearly as stable as v6 and I don't really know why at this point. Some of this is likely breadboard wiring but it sure does seem way more sensitive to that now as the best I could run it was 1/2 the clock speed compared to v6 demos.

I will also take a look at spi transfers to see if I changed anything not meant to be changed, will update this PRs with more fixes.

Regards

C47D:
With 8ba9d9d, only the source code of the display enabled on the menuconfig is added into the build, this might help us to remove the `disp_driver_init` and `disp_driver_flush` functions on the `disp_driver.c` file and by renaming the display controller init function, example `ili9341_init` into `disp_driver_init` we can call the display init function directly, so we avoid the `disp_driver` layer.

What do you think?

davidjade:
@C47D 

I cannot build the 98ad8aed97eaf6719386aaa560d1d915c4d9506e Dev-7.0 branch because now when selecting the FT81X for TFT and Touch because neither CONFIG_LVGL_TOUCH_CONTROLLER_SPI_HSPI or CONFIG_LVGL_TOUCH_CONTROLLER_SPI_VSPI are getting defined, which causes TOUCH_SPI_HOST to not get defined (and I'm not even sure it should since there is no separate touch controller code). I'm not even sure where they would get defined since these two don't appear in any kconfig file. 

This ultimately causes a call to tp_spi_add_device() for the  FT81X which is wrong - the SPI bus is already set up for the display and nothing needs to be done to add a separate touch controller like other displays. The FT81X is unique is this regard - it is integrated in one device and only has one SPI bus/channel for both display and touch. This seems like maybe a new case for the lvgl_helpers that sets up the SPI devices - just set up one device for the FT81X (the display) and that's all because the display driver handles touch - not a separate touch driver.

I am not sure what the best option for config is long term. Right now, I can config the FT81X for TFT and also config the FT81X for touch and that seems right from the menuconfig perspective. It allows to disable touch if you want to. But when they are both selected, something needs to stop the pragma/error that no touch controller is selected on the FT81X because one is - the display integrated one. Config also needs to stop any other code from being called that tries to set up a separate touch controller on the SPI bus because there is no separate touch controller for the FT81X.

C47D:
Now that I think of it, you shouldn't be able to choose the SPI master for FT81X touch controller, because, as you said, it's the same used for the display. Let me know if the last two commits improve your use case.

If you think is best we can hide the "Choose SPI Master port" option when selecting the FT81X on the touch menu and handle it so you don't get the "No SPI master for touch" issue.

davidjade:
Thanks. Things build now and run.

I agree that it does not make much sense to select a SPI bus for the touch controller since it is using the unified FT81X driver. Maybe just the ability to select/deselect the FT81X touch driver to enable/disable touch. 

It still only runs at a slower clock speed for some unknown reason. It is odd I can flip back and forth between V6 & V7 demo projects and run at twice the clock speed with V6 on the same piece of hardware. I have no idea why but it could come down to ESP32 kconfig settings between the two projects or some other timing issue so I am looking into that. The setup code path really looks the same but clearly something is happening somewhere, maybe just a timing issue. 

One other thought though: these #pragma statement for which drivers are init'd, etc.. - are they going to be permanent? They create a lot of noise in the build. I get them for logging but during building I hope they aren't a permanent code feature. "FT81X Display and Touch enabled" in particular gets spit out 8 times, each creating a 7-line block in the output.

C47D:
> One other thought though: these #pragma statement for which drivers are init'd, etc.. - are they going to be permanent? They create a lot of noise in the build. I get them for logging but during building I hope they aren't a permanent code feature. "FT81X Display and Touch enabled" in particular gets spit out 8 times, each creating a 7-line block in the output.

They are not, I've been using them to know what symbols are being defined and which not, I'm replacing them with ESP_LOGI where possible.

davidjade:
While trying to understand what may be causing the new V7 version to need slower clock speeds with the DT81X, I tried to build with Platform.io the same as I usually do with my V6 FT81X code. I found one issue: 

The _set(EXTRA_COMPONENT_DIRS..._ vs. _list(APPEND EXTRA_COMPONENT_DIRS..._ choice in the top level CMakeLists.txt file. PIO needs the list(APPEND syntax or the build fails. I believe this is because they parse things separately to try and handle certain things themselves and if you re-set the list instead of appending, then a bunch of stuff gets left out and the build fails. I would suggest you switch to the _list(APPEND_ syntax as I believe that will work everywhere and is maybe more correct since other EXTRA_COMPONENT_DIRS list items may already exist that need to be preserved instead of replacing the whole list.

C47D:
I've tried it but couldn't make it work, should the list(APPEND ...) CMake command needs to be added to the lvgl_tft, lvgl_touch, lvgl and lv_examples CMakeLists.txt files, isn't? IIRC you had it working, can you send a PR against this branch with the working code?

If we switch to the proposed separated lvgl, lv_examples and lv_esp32_drivers repos we will not need to set the EXTRA_COMPONENT_DIRS.

davidjade:
Submitted as #138 

I have verified that it works for both ESP-IDF and PlatformIO build systems using the v4.0 versions of each (i.e. one small step behind recent releases).

C47D:
Thanks, have you been able to find the reason for the slow down on your project? I will have some time to work on this later today.

davidjade:
Not yet but I tried something last night  to pin it down. I moved all the new driver code to the V6 version as it is compatible (with one small symbol name change). This brought the slow down/corruption issue to my V6 based projects, so something is up with the new code but I have scoured it and have not seen any differences that could explain it.

C47D:
The difference might be on the LVGL configuration, or the SPI Clock.

davidjade:
What I tried to do was isolate just the new driver code from the old. I took the stock V6 demo project, the same one I used to develop the FT81x drivers with, and replaced just the lvgl_esp32_drivers code folder. So it's the same v6 LVGL version, ESP32 and SPI settings, base sdkconfig file, etc... It should behave the same but it does not. It is only stable at 1/2 the SPI clock speed as the my v6 SPI code that got checked into master. I behaves like bad SPI wiring but that's not it because I can flip back and forth old vs. new code on the same hardware and see the problem.

As I mentioned I have really looked for differences that could account for this in the new SPI code but I have yet to find anything. Starting to wonder if there isn't something else going on - some sort of memory corruption or contention issue, but I haven't found it so far. To me it looks like it should be very close to the old code execution path despite mostly just moving around the SPI config code. The only real different is the new logging stuff but even if I remove that, it's still not working right.

I do have a logic analyzer here but I haven't tried that yet. That might be the next step and maybe it will point to the difference that is causing the instability.

davidjade:
Progress today on the clock speed issues. In short, I don't think what I am seeing is strictly a software issue.

For sure the V6 and V7 driver code using LVGL v6 behaves differently on the same hardware at clock speeds greater than 16Mhz, despite being essentially the same code - just rearranged. However I don't think the issue comes from something wrong with the new code. 

Today I put a logic analyzer on it and I can see the same bits getting pushed over the wire but what is really telling is, just having the logic analyzer plugged into the circuit seems to make the issue better or even sometimes completely go away. Even if the analyzer is powered down, it works better.

So without the logic analyzer in circuit I can only surmise that the SPI data is getting corrupted as the display just shows junk at higher clock speeds. This probably explains why just slowing down the clock makes it go away. 

Why there is different behavior between the old and new code is still a mystery but I am pretty sure at this point it is something more than just the new code. Perhaps the rearrangement of code has just caused some timing change that uncovered a problem with the hardware I use to test. 

I am not surprised that it appears to be a more of a wiring issue with higher clock speeds but I am really surprised that a change in code or rather just a rearrangement of code, can make it disappear/reappear.

At this point though, the new code seems to work on my FT81x - at least with a logic analyzer plugged in.

C47D:
That is weird, indeed, do you think the pins are being initialized incorrectly? I've been working with a ST7789 based display and had not seen slowdowns when using this branch.

davidjade:
From my reading of the spi_master.c source, it handles all GPIO pin setup so I don't think that is the issue. It does not however set an internal pull-up config (which I think would only be a weak pull-up). So I suspect that it is just my particular display hardware, being on a breadboard, etc.. that fouls this up in certain situations at higher clock speeds. 

I've got another display here that I can try but I have to make a wiring harness for it. I should be able to better control the wire length though with that and the pull-up situation better. So I'll try that and see what happens.

C47D:
@vheathen Are you able to sign the CLA agreement?

vheathen:
> @vheathen Are you able to sign the CLA agreement?

Sure, done!

C47D:
I'm having issues with a SSD1306 display, the flush function isn't being called, investigating...

an-erd:
When using the example in v7.0, I get not the text shown but more like the surrounding box of that text, in the image the text "Hello world!" is shown. Same error for the Wemos Lolin (ssd1306) and M5Stick (SH1107). Everything works correctly with v6.

Example GUI with LittlevGL on ESP32

Similar issue as in [Lv_task_handler returns 4294967295](https://forum.lvgl.io/t/lv-task-handler-returns-4294967295/2382), where I put this issue, too.

kisvegabor:
Do you take the `opa` parameter into account in `set_px_cb`?

C47D:
@kisvegabor This is the current `set_px_cb` for the SSD1306 display

https://github.com/lvgl/lv_port_esp32/blob/0739f5b61f0b0388b97d3574af4929d136b46794/components/lvgl_esp32_drivers/lvgl_tft/ssd1306.c#L133-L143

C47D:
New `set_px_cb` function:

```c
void ssd1306_set_px_cb(struct _disp_drv_t * disp_drv, uint8_t * buf, lv_coord_t buf_w, lv_coord_t x, lv_coord_t y,                                                  134         lv_color_t color, lv_opa_t opa) {
    uint16_t byte_index = x + (( y>>3 ) * buf_w);
    uint8_t  bit_index  = y & 0x7;

    if ((color.full == 0) && (LV_OPA_TRANSP != opa)) {
        BIT_SET(buf[byte_index], bit_index);
    } else {
        BIT_CLEAR(buf[byte_index], bit_index);
    }
}
```

Output:
WhatsApp Image 2020-06-02 at 18 41 15

So I think it's solved, at least for the SSD1306, @an-erd could you test it for your other monochrome display?

an-erd:
@C47D Yes, this resolves the issue. I opend an PR for the SH1107, too. Thanks!

C47D:
@kisvegabor @embeddedt This seems ready to be merged into master (conflicts will be resolved when merging), I would not like to delay more this PR, it ended up solving a lot of issues apart of being port to LVGL v7.

components/lvgl_esp32_drivers/lvgl_tft/disp_spi.c
```diff
@@ -178,9 +166,22 @@ bool disp_spi_is_busy(void)
 
 static void IRAM_ATTR spi_ready (spi_transaction_t *trans)
 {
-    spi_trans_in_progress = false;
+    disp_spi_send_flag_t flags = (disp_spi_send_flag_t) trans->user;
+
+    if (flags & DISP_SPI_SIGNAL_FLUSH) {
+        lv_disp_t * disp = NULL;
+
+#if LV_VERSION_CHECK(7,0,0)
```
embeddedt:
```suggestion
#if LVGL_VERSION_MAJOR >= 7
```

There aren't any plans to drop the `_` naming convention and the current condition will break in 8.0.

components/lvgl_esp32_drivers/lvgl_tft/ili9488.c
```diff
@@ -121,7 +117,11 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
     uint32_t size = lv_area_get_width(area) * lv_area_get_height(area);
 
     lv_color16_t *buffer_16bit = (lv_color16_t *) color_map;
-    uint8_t *mybuf = (uint8_t *) heap_caps_malloc(3 * size * sizeof(uint8_t), MALLOC_CAP_DMA);
+    uint8_t *mybuf;
+    do {
+        mybuf = (uint8_t *) heap_caps_malloc(3 * size * sizeof(uint8_t), MALLOC_CAP_DMA);
+        if (mybuf == NULL)  ESP_LOGW(TAG, "Could not allocate enough DMA memory!");
+    } while (mybuf == NULL);
```
embeddedt, line 124: How would more memory become available in a realistic scenario? This looks like it could end up spinning in a loop indefinitely. Perhaps an `abort()` or the ESP equivalent should be used instead?

components/lvgl_esp32_drivers/lvgl_tft/sh1107.c
```diff
@@ -165,14 +161,25 @@ void sh1107_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * colo
 #else 
         ptr = color_map + i * CONFIG_LVGL_DISPLAY_WIDTH;
 #endif
-        sh1107_send_color( (void *) ptr, size);
+        if(i != row2){
+			sh1107_send_data( (void *) ptr, size);
+		} else {
+			// complete sending data by sh1107_send_color() and thus call lv_flush_ready()
+			sh1107_send_color( (void *) ptr, size);
+		}
```
embeddedt: I'm generally not picky about indentation but this is confusing.

main/main.c
```diff
@@ -84,9 +99,14 @@ void guiTask() {
...
-    lv_theme_mono_init(0, NULL);
-    lv_theme_set_current( lv_theme_get_mono() );
+#ifdef CONFIG_LVGL_TFT_DISPLAY_MONOCHROME
+    lv_theme_mono_init(LV_THEME_DEFAULT_COLOR_PRIMARY, LV_THEME_DEFAULT_COLOR_SECONDARY,
+		LV_THEME_DEFAULT_FLAG,
+		LV_THEME_DEFAULT_FONT_SMALL,
+		LV_THEME_DEFAULT_FONT_NORMAL,
+		LV_THEME_DEFAULT_FONT_SUBTITLE,
+		LV_THEME_DEFAULT_FONT_TITLE);
+    lv_theme_set_act( lv_theme_get_act() );
```
embeddedt: This doesn't do anything (take a look at the [implementation](https://github.com/lvgl/lvgl/blob/950ac3b9930ed62bf86fe411b94f234ae644da4d/src/lv_themes/lv_theme.c#L42-L54)).
```suggestion
```

C47D: Thanks for taking the time to review it, I just saw the `lv_init` implementation and this step (configuring the theme) is already done in there:

https://github.com/lvgl/lvgl/blob/950ac3b9930ed62bf86fe411b94f234ae644da4d/src/lv_core/lv_obj.c#L144-L147

So I might take this whole block of code away:

https://github.com/lvgl/lv_port_esp32/blob/3f922ee95e11d54fee2f631b0cb32b8f10a1e103/main/main.c#L102-L110

components/lvgl_esp32_drivers/lvgl_tft/sh1107.c
C47D: I will fix that. Didn't noticed it. Fixed with da04ed1, I will need to add a code style checker as LVGL does.

components/lvgl_esp32_drivers/lvgl_tft/disp_spi.c
C47D: Thanks, I will fix that, fixed in ff994aa, on the LVGL configuration files (Kconfig and lv_conf) I was using `LV_`, will update those to `LVGL_`.

CONTRIBUTE_CONTROLLER_SUPPORT.md
```diff
@@ -0,0 +1,110 @@
+# Is your driver not listed on the supported drivers list?
+
+Consider contribute to this demo project by adding support to your driver.
```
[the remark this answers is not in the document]
C47D: Fixed with 2e250af.

main/main.c
C47D: Removed in a526922

components/lvgl_esp32_drivers/lvgl_tft/ili9488.c
C47D: The SPI code was reworked because @davidjade detected some potential memory leaks #112, may be the `do {} while` loop is not longer needed, I will test it later with a ILI9488 display.

davidjade: I agree that I don't see how this would really work once it fails to allocate unless there is some way to free DMA capable memory to retry. But it is the right thing to do, to allocate DMA capable memory in anything that gets handed off to a SPI request. 

This gets really trick though as that allocated DMA memory is now tied up in a SPI transaction and cannot be freed until the transaction results are retrieved/completed. I did not look into this particular driver so I don't know if it does that. If it doesn't, then it likely has to be freed in the SPI interrupt handler or some other specific driver callback upon SPI transaction completion. Or maybe held until the next time and reused/resized.

It would probably be best to not do any DMA allocations if possible and use a static buffer like the other drivers (or maybe just allocate once and reuse it).

C47D: Thanks, I will look into using a statically allocated memory chunk.

components/lv_examples/component.mk
```diff
@@ -4,24 +4,12 @@
...
-    lv_examples/lv_tutorial/8_animations    \
-    lv_examples/lv_tutorial/9_responsive
-
+COMPONENT_SRCDIRS := lv_examples           \
+    lv_examples/src/lv_demo_benchmark      \
+    lv_examples/src/lv_demo_keypad_encoder \
+    lv_examples/src/demo_stress            \
+    lv_examples/src/lv_demo_widgets        \
+    lv_examples/src/lv_ex_style            \
+    lv_examples/src/lv_ex_widgets
```
kisvegabor: [lv_examples/assets](https://github.com/lvgl/lv_examples/tree/master/assets) might be also required.

kisvegabor:
Amazing work! I found only one thing, see the comment.

At some point, it'd be great to move these drivers to `lv_driver` to make available in other projects. To do this the hardware API of `lv_drivers` needs to be updated to provide the required abstraction.
But it's another story, just mentioning it... :slightly_smiling_face:

davidjade:
@C47D I am getting compile errors in master now. Seems to be two symbols when I select the FT81x that prevents building: CONFIG_LVGL_TOUCH_SPI_MISO & TP_SPI_MISO


I get the errors in a bunch of files:

```
In file included from ../components/lvgl_esp32_drivers/lvgl_helpers.h:17,
                 from ../components/lvgl_esp32_drivers/lvgl_helpers.c:10:
../components/lvgl_esp32_drivers/lvgl_helpers.c: In function 'lvgl_driver_init':
../components/lvgl_esp32_drivers/lvgl_spi_conf.h:36:21: error: 'CONFIG_LVGL_TOUCH_SPI_MISO' undeclared (first use in this function); did you mean 'CONFIG_LVGL_DISP_SPI_MISO'?
 #define TP_SPI_MISO CONFIG_LVGL_TOUCH_SPI_MISO
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~
../components/lvgl_esp32_drivers/lvgl_helpers.c:77:9: note: in expansion of macro 'TP_SPI_MISO'
         TP_SPI_MISO, DISP_SPI_MOSI, DISP_SPI_CLK,
         ^~~~~~~~~~~
../components/lvgl_esp32_drivers/lvgl_spi_conf.h:36:21: note: each undeclared identifier is reported only once for each function it appears in
 #define TP_SPI_MISO CONFIG_LVGL_TOUCH_SPI_MISO
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~
../components/lvgl_esp32_drivers/lvgl_helpers.c:77:9: note: in expansion of macro 'TP_SPI_MISO'
         TP_SPI_MISO, DISP_SPI_MOSI, DISP_SPI_CLK,
         ^~~~~~~~~~~
../components/lvgl_esp32_drivers/lvgl_spi_conf.h:35:21: error: 'CONFIG_LVGL_TOUCH_SPI_MOSI' undeclared (first use in this function); did you mean 'CONFIG_LVGL_DISP_SPI_MOSI'?
 #define TP_SPI_MOSI CONFIG_LVGL_TOUCH_SPI_MOSI
                     ^~~~~~~~~~~~~~~~~~~~~~~~~~
../components/lvgl_esp32_drivers/lvgl_helpers.c:133:26: note: in expansion of macro 'TP_SPI_MOSI'
             TP_SPI_MISO, TP_SPI_MOSI, TP_SPI_CLK,
                          ^~~~~~~~~~~
../components/lvgl_esp32_drivers/lvgl_spi_conf.h:37:21: error: 'CONFIG_LVGL_TOUCH_SPI_CLK' undeclared (first use in this function); did you mean 'CONFIG_LVGL_DISP_SPI_CLK'?
 #define TP_SPI_CLK  CONFIG_LVGL_TOUCH_SPI_CLK
                     ^~~~~~~~~~~~~~~~~~~~~~~~~
../components/lvgl_esp32_drivers/lvgl_helpers.c:133:39: note: in expansion of macro 'TP_SPI_CLK'
             TP_SPI_MISO, TP_SPI_MOSI, TP_SPI_CLK,
                                       ^~~~~~~~~~
```

Seems to be related to the FT18x not needing separate TOUCH config again. It also looks like maybe a touch driver is trying to initialized SPI again in lvgl_helpers.c when the FT81x is selected.

C47D:
Will take a look at it

Issue #119, opened 2020-05-08
Colors are inverted on ILI9341


rangeli:
We use GitHub issues for development related discussions.
Please use the [forum](https://forum.littlevgl.com/) to ask questions.

**Describe the issue**
I flashed the example to my [M5Stack Gray](https://docs.m5stack.com/#/en/core/gray) and the colors were inverted.

I tried using the "Invert display" feature, but found that it actually flipped the display instead.

I was able to fix this by sending an invert display command to the ILI9341, but I had to modify the code to do this.

I've created a PR with the change, but I'm not sure if this issue only affects some M5Stack models.

**Code to reproduce the issue**

**Expected Results**

**Actual Results**

**ESP32 Chip version**

**ESP-IDF version**

**Development kit used**

**Development machine OS**

**Compilation warnings/errors (if available)**

**If possible, copy the compilation log into a file and attach it here**

C47D:
Hi,

I don't have that dev kit, can you take a pic of the inverted colors? IIRC I had the same problem but with other display.

Based on your other issue, we may add an option to the display menu config named "Invert display colors" or something similar, so the users can enable/disable at will. What do you think?

rangeli:
I think having a menu item for this might make sense, at least for the M5Stack/ILI9341.  I've made a PR with the fix BTW, but I could turn it into a menuconfig option.  Here's some pics of the display before and after the fix:
![image](https://user-images.githubusercontent.com/1470630/81382423-60d50700-90dc-11ea-8575-94eff5c645d6.png)

![image](https://user-images.githubusercontent.com/1470630/81382541-8feb7880-90dc-11ea-85f5-95ccc03f2897.png)

C47D:
Thanks for the photo and PR, the problem can be specific to the M5Stack, I'm using the Wrover kit for my tests and I didn't got the inverted colors.

By any chance are you configuring the display orientation to any of the inverted options?

barbiani:
Try to change your lv_conf.h to `#define LV_COLOR_16_SWAP   1`

rangeli:
> Try changing your lv_conf.h to `#define LV_COLOR_16_SWAP 1`

Unfortunately, that doesn't work.  That option swaps the the 2 bytes of RGB565 color.  The fix needs to be a full invert of the colors.

rangeli:
I've added a menuconfig option for this:

[...]

PR #121, opened 2020-05-08
Add menuconfig option to invert colors on ILI9341

rangeli:
This fixes issue #119


rangeli:
@C47D

components/lvgl_esp32_drivers/lvgl_tft/ili9341.h
```diff
@@ -40,6 +40,7 @@ extern "C" {
 #endif
 
 #define ILI9341_INVERT_DISPLAY CONFIG_LVGL_INVERT_DISPLAY
+#define ILI9341_INVERT_COLORS LVGL_INVERT_COLORS
```
C47D: `LVGL_INVERT_COLORS` must be `CONFIG_LVGL_INVERT_COLORS`, the `CONFIG` prefix is added.

components/lvgl_esp32_drivers/lvgl_tft/ili9341.c
```diff
@@ -159,6 +159,12 @@ void ili9341_init(void)
 	ili9341_send_cmd(0x36);
 	ili9341_send_data(&data, 1);
 #endif
+
+#if defined ILI9341_INVERT_COLORS
```
C47D: Check `build/config/sdkconfig.h` to know what value `CONFIG_LVGL_INVERT_COLORS` have if you set to option to false, I don't know if the option is set to 0 or simply not defined.

If it is set to 0 when you don't enable the option then I think this must be:
```c
#if ILI9341_INVERT_COLORS == 1
```

components/lvgl_esp32_drivers/lvgl_tft/ili9341.h
mica-angeli: Fixed

components/lvgl_esp32_drivers/lvgl_tft/ili9341.c
mica-angeli: The `CONFIG_LVGL_INVERT_COLORS` value equals `1` when it's set and it's undefined when not set.

I changed it to check the value and it's working correctly both when the color inversion option is set and when it is not set.

C47D:
Thanks for the PR, I can't test it right now, so will test it later today.

Issue #120, opened 2020-05-08
M5Stack display is scrambled


rangeli:
We use GitHub issues for development related discussions.
Please use the [forum](https://forum.littlevgl.com/) to ask questions.

**Describe the issue**
The display is scrambled on the latest version on the M5Stack configuration.  I checked out every previous PR until I found one that worked for me.  The PR that caused the bug was #113.

@C47D 

**Code to reproduce the issue**

**Expected Results**

**Actual Results**

**ESP32 Chip version**

**ESP-IDF version**

**Development kit used**

**Development machine OS**

**Compilation warnings/errors (if available)**

**If possible, copy the compilation log into a file and attach it here**

PR #122, opened 2020-05-08
Fix issues with scrambled M5Stack screen

rangeli:
Not sure what the old code was trying to do, but this fixed issue #120

### DIFF
--- a/components/lvgl/lv_conf.h
+++ b/components/lvgl/lv_conf.h
@@ -23,19 +23,8 @@
  *====================*/
 
 /* Maximal horizontal and vertical resolution to support by the library.*/
-#if defined (CONFIG_LVGL_PREDEFINED_DISPLAY_NONE) && defined (CONFIG_LVGL_DISPLAY_ORIENTATION_PORTRAIT)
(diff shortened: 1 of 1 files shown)


C47D:
The old code handles when the user doesn't choose a predefined configuration and set up a rotation for the display, I guess you were using the M5Stack configuring every option for yourself.

Right now the only "validated" options for the M5Stack (when choosing the predefined configuration) are listed below:
https://github.com/littlevgl/lv_port_esp32/blob/32c4ab0f734b0d6afb1b444633aaba6b8e3c0320/components/lvgl_esp32_drivers/lvgl_tft/ili9341.c#L119-L131

If you set everything for yourself the values to configure the MADCTL register are the following:
https://github.com/littlevgl/lv_port_esp32/blob/32c4ab0f734b0d6afb1b444633aaba6b8e3c0320/components/lvgl_esp32_drivers/lvgl_tft/ili9341.c#L147-L166

You can tell the values are different so your M5Stack behaves wrong.

Can you confirm or not if you are not using the predefined configuration?

C47D:
@rangeli Sorry, i just pushed a PR, can you test #124?

PR #126, opened 2020-05-08
Change SPI clock frequency from menuconfig, ILI9488 default 40MHz.

persello:
- Added ability to change SPI clock frequency from menuconfig.
- Set default frequency for ILI9488 to 40MHz (as per datasheet).


C47D:
It looks very useful, do you want me to merge it right away?

persello:
As pointed out in the issue, the ESP32 does not have fractional multipliers for the SPI clock, so a multiple choice menu could be a better solution. I'm going to read the ESP32 documentation to better understand what choices we have and do some tests tonight with a different menu setup.

persello:
OK, it seems that the possible frequencies are obtained dividing the APB clock by an integer number from 1 to 10. I didn't find any related setting in menuconfig, but it seems that the APB is 80MHz, obtained dividing the CPU clock by three when set to 240MHz and by two when set to 160MHz. Although the APB clock could be changed, it seems to be meant to be always at 80MHz, from what I can read in some tables regarding the power management of the ESP32.

I'm not an expert in this field, but I would go with a menu from which you can select one of the ten available frequencies (rounded to the megahertz, letting ESP-IDF choose the correct divider), instead of letting the user freely choose it and then maybe get an unexpected result. For the default frequencies of the displays, I would round them to the nearest value, except for those that would round to an excessively high value.

persello:
Updated. I'll mark it ready for review.

components/lvgl_esp32_drivers/lvgl_spi_conf.h
```diff
@@ -70,22 +70,19 @@ extern "C" {
...
+#define SPI_TFT_CLOCK_SPEED_HZ  (20*1000*1000)
 #elif defined CONFIG_LVGL_TFT_DISPLAY_CONTROLLER_HX8357
 #define SPI_TFT_CLOCK_SPEED_HZ  (26*1000*1000)
 #elif defined CONFIG_LVGL_TFT_DISPLAY_CONTROLLER_SH1107
 #define SPI_TFT_CLOCK_SPEED_HZ  (8*1000*1000)
 #elif defined CONFIG_LVGL_TFT_DISPLAY_CONTROLLER_ILI9486
-#define SPI_TFT_CLOCK_SPEED_HZ  (24*1000*1000)
-#elif defined CONFIG_LVGL_TFT_DISPLAY_CONTROLLER_ILI9488
-#define SPI_TFT_CLOCK_SPEED_HZ  (40*1000*1000)
+#define SPI_TFT_CLOCK_SPEED_HZ  (20*1000*1000)
```
C47D, line 83: Is 40MHz a reliable default clock for the ILI9486?

components/lvgl_esp32_drivers/lvgl_spi_conf.h
```diff
@@ -70,22 +70,19 @@ extern "C" {
 #define SPI_BUS_MAX_TRANSFER_SZ (DISP_BUF_SIZE * 2)
 #endif
 
-#if defined (CONFIG_LVGL_TFT_USE_CUSTOM_SPI_FREQUENCY)
-#define SPI_TFT_CLOCK_SPEED_HZ (CONFIG_LVGL_TFT_SPI_CUSTOM_FREQUENCY * 1000 * 1000)
+#if defined (CONFIG_LVGL_TFT_USE_CUSTOM_SPI_CLK_DIVIDER)
+#define SPI_TFT_CLOCK_SPEED_HZ (80 * 1000 * 1000 / CONFIG_LVGL_TFT_CUSTOM_SPI_CLK_DIVIDER)
```
C47D: I think the multiplication operator have precedence over the divisor operator [1], but I think is better to use parenthesis for it.
```c
#define SPI_TFT_CLOCK_SPEED_HZ ((80 * 1000 * 1000) / CONFIG_LVGL_TFT_CUSTOM_SPI_CLK_DIVIDER)
```

[1] https://en.cppreference.com/w/c/language/operator_precedence

components/lvgl_esp32_drivers/lvgl_spi_conf.h
C47D: Also, could you add an 'implicit' default value for the ILI9341 and ILI9488 displays?

persello: Sure, right now I’m not at the PC anymore, tomorrow I will take a look. Regarding ILI9488, 40MHz is perfectly working even on a breadboard.

Edit: I read ILI9488, sorry! I have no knowledge about that controller. Will check its data sheet tomorrow.

C47D: Thanks, I think 40MHz for all ILI9* controllers is ok on breadboards, I'm wiring my ILI9488 that way, maybe I'm just lucky :)

persello: Ok, reading some comments on the ILI9486, it seems that hardly reaches values above 30MHz, so until someone could test it, I would leave it to 20MHz (rounded from the 24MHz you set as default before).

C47D:
Thanks for the PR @persello, dev 7 is getting almost ready to be merged.

Issue #112, opened 2020-05-05
TFT SPI queued but never dequeued - can lead to potential memory leaks


davidjade:
Sorry for the long narrative but this doesn't fit the bug template as this is a general design issue with the current disp_spi driver.

I've been working with the lvgl esp32 driver code (adding a new display type actually) and something strikes me as wrong.

Code in the TFT disp_spi calls spi_device_queue_trans() but it never calls spi_device_get_trans_result(). From reading the IDF spi_master source, this places items into an xQueue for received transactions but leaves them there. This causes the spi_master received queue to overflow silently because spi_master does not check for that. I guess if you never need results, this is maybe somewhat Ok. However, more worrisome is that sometimes the spi_master will allocate and attach DMA memory to a transaction (if the transaction buffers passed in are in non DMA-able memory) and this only gets freed if you retrieve the transaction results (via uninstall_priv_desc()).

There is a note in spi_master.c about this (in spi_device_get_trans_result(), around line 935): 
```
// The memory occupied by rx and tx DMA buffer destroyed only when receiving from the queue (transaction finished).
```
So if you have conditions that cause spi_master to allocate DMA memory and you never retrieve the SPI transaction results, then that memory never gets freed.

Maybe no one has seen this since a lot of folks statically allocate the display buffers and that ends up in IRAM. In my case I have a 800x480 16-bit screen device and I trying to see how it would perform using PSRAM in some cases, which is not DMA-able as far as I know. With the current lvgl_esp32_driver design, these internally allocation spi_master DMA buffers will never get freed and will lead to memory leaks.

I think the prudent thing to do is to always wait for and retrieve transaction results before sending new transactions, so that both the spi_master receive xQueue doesn't overflow and any internally attached DMA memory gets freed. This likely will cause how spi_trans_in_progress is signaled to have to change too as it would then be more complicated than just  post_cb being called as a transaction really isn't finished until the results are retrieved.

C47D:
Hi @davidjade thanks for reaching out.

[...]

PR #130, opened 2020-05-17
Rework SPI to properly handle SPI transactions (#112) and add FT81x driver

davidjade:
This addresses issue #112 by handling transactions results in a way that allows any internal SPI driver allocations to be freed. It also prevents the ESP32's spi_master component/driver queue from overflowing.

In addition, the reworked SPI driver has expanded capabilities to enable a new driver for the FT81x series of touchscreen displays. These changes maintain compatibility with existing drivers with one small change to them, which should provide multitasking benefits on its own. 

A few notes on the FT81x series:

1. The FT81x display are co-processor displays so a bit more complicated to use. They require writing and reading SPI transactions for instance. The FT81x driver is based on a fork of an MIT license FT81x driver, which I have heavily modified to suit LvGL. For this new driver, it gets treated as a simple memory-mapped display similar to other TFTs.

2. The full FT81x capabilities are included in the support code but if-def-ed out. They can still be enabled to allow for a hybrid display. You could for instance tell LvGL that only part of the display was to be used leaving the other areas to the FT81x co-processor or even implementing overlays, video, etc... using the co-processor. You could also implement split-screens by setting up two or more regions as "separate" LvGL displays.

3. The max SPI speed of the FT81x is 30Mhz and they are typically 800x480 resolution so they are not particularly fast for full screen updates due to the amount of data (when not fully using the co-processor for native graphics) but they are still usable with LvGL. I will likely work on enabling Quad SPI as a future driver enhancement which should bring the frame rate up to 14-16 full FPS for 800x480 16-bit displays.

4. The FT81x supports several pixel formats natively so you could configure it as a RGB565 (current config), RGB332, RGB with alpha, grayscale, or monochrome display with LvGL. This would require some additional changes to the FT81x config code, but it should work.


C47D:
Hi @davidjade,
Thanks for the PR, I gave it a quick check and everything seems correct, I will do some tests with the displays I have at hand.

The LVGL v7 will be released soon, so I want to merge this into master and then create a separated branch for it, so we can get the v7 PR merged into master, also we can try to reuse as much code as possible from this PR. It will be done manually because there was some changes already on the display handling.

davidjade:
Ok, I can do a re-check on my FT81x displays when it's merged with V7 to make sure everything is working Ok, etc... I'm also happy to look over any manual merges too, if that helps. Let me know when you need that check or have any other questions.

components/lvgl_esp32_drivers/lvgl_tft/Kconfig
```diff
@@ -167,11 +174,16 @@ menu "LittlevGL (LVGL) TFT Display controller"
...
+	config LVGL_TFT_DISPLAY_USER_CONTROLLER_SSD1306
+		bool "SSD1306"
+			select LVGL_TFT_DISPLAY_CONTROLLER_SSD1306
+			select LVGL_TFT_DISPLAY_PROTOCOL_I2C
+			select LVGL_TFT_DISPLAY_MONOCHROME
+	config LVGL_TFT_DISPLAY_USER_CONTROLLER_FT81X
+		bool "FT81X"
+			select LVGL_TFT_DISPLAY_CONTROLLER_FT81X
+			select LVGL_TFT_DISPLAY_PROTOCOL_SPI
+			select LVGL_DISPLAY_ORIENTATION_LANDSCAPE
```
C47D: Can you remove this line? `select LVGL_DISPLAY_ORIENTATION_LANDSCAPE`, that is the only warning I'm getting so far.

C47D:
I'm also testing a new way to add LVGL to custom projects, this is proof of concept, it it works i hope it gets integrated in the LVGL repository.

This is the LVGL component: https://github.com/C47D/lv_esp32_component
This is my test project: https://github.com/C47D/lv_esp32_myproject

Can you check it if you have some free time? I'm planning to update the lv_config.h and lv_ex_config.h this days, also add a Kconfig to the examples directory to be able to choose which example to run or not run an example at all.

components/lvgl_esp32_drivers/lvgl_tft/Kconfig
davidjade: Ok, I have updated my fork to removed that. Sorry about that, it was something that should have been removed.

davidjade:
I will try and look into your new stuff more this weekend but here's my quick take.

It sure would be nice though to have a structure that did not require adding to EXTRA_COMPONENT_DIRS. It's the too many extra levels of folders that prevents that as I am sure you are aware. I would actually just prefer separate component repos for lvg, esp32 drivers, and examples (if I want them) - do away with the subrepo dependancies. I'm fine with  (and would prefer to) managing my own repo dependencies.

btw, in my local test projects I use this so that I don't clobber the extra component dirs but rather add to them. I had to do this because it was messing with some other ESP-IDF native component projects. Append vs Set:

list(APPEND EXTRA_COMPONENT_DIRS ...)

Another thing I had to do is specify an earlier version of cmake as I think the IDF SDK ships with a different version than PIO and I want projects to be buildable with standard ESP-IDF projects too.

cmake_minimum_required(VERSION 3.13.0)

C47D:
I will merge this into master so it can be tested for others, I haven't been able to get an oscilloscope or a logic analyzer fast enough to see how busy the SPI master gets or if it gets corrupted.

---

> btw, in my local test projects I use this so that I don't clobber the extra component dirs but rather 
> add to them. I had to do this because it was messing with some other ESP-IDF native component projects. Append vs Set:
> 
> list(APPEND EXTRA_COMPONENT_DIRS ...)

I will try this, the other repo can be changed to make it easier to use.

> It sure would be nice though to have a structure that did not require adding to EXTRA_COMPONENT_DIRS. It's the too many extra levels of folders that prevents that as I am sure you are aware. I would actually just prefer separate component repos for lvg, esp32 drivers, and examples (if I want them) - do away with the subrepo dependancies. I'm fine with (and would prefer to) managing my own repo dependencies.

I was actually trying to figure out how to add a Kconfig in the lv_examples repo, so the user can choose to run a given example or not at all.

> Another thing I had to do is specify an earlier version of cmake as I think the IDF SDK ships with a different version than PIO and I want projects to be buildable with standard ESP-IDF projects too.
> 
> cmake_minimum_required(VERSION 3.13.0)

Will change that.

PR #138, opened 2020-05-28
Updated CMakeLists.txt for PlatformIO

davidjade:
This updates CMakeLists.txt for be compatible for both ESP-IDF and PlatformIO build systems


C47D:
Thanks @davidjade, I will test it on my machine later today. I will need to update the README.

PR #139, opened 2020-05-29
Fix monochrome displays using ssd1306 and sh1107

an-erd:
- Workaround using ssd1306_rounder() and sh1107_rounder() for ssd1306_flush() and sh1107_flush() which needs to be fixed.
- Fixed: lv_flush_ready() is called to early for sh1107
- Remove including lv_theme_mono.h


C47D:
Hi @an-erd,

Could you make the PR against dev-7.0? After checking the code I think it should not require additional changes.

PR #143, opened 2020-06-02
Fix example (with two screen-sized buffers and set_px_cb)

an-erd:
Fix the example because two screen-sized buffers cannot be used with set_px_cb callback set.


C47D:
Thanks @an-erd !

PR #144, opened 2020-06-03
SH1107: Check for opa when setting a pixel on set_px_cb

an-erd:


kisvegabor:
I've fixed in lvgl too. Now `set_px_cb` is not called when the pixel is transparent.

C47D:
Thanks @an-erd and @kisvegabor!

PR #153, opened 2020-06-09
lv_printf: expose floating-point configuration option to menuconfig

an-erd:
Fix: With v7 the option to configure disabling float in lv_prinf functions is exposed, but not put in lv_conf.h. The option can now be changed with menuconfig without modifying lv_conf.h.

### DIFF
--- a/components/lvgl/lv_conf.h
+++ b/components/lvgl/lv_conf.h
@@ -833,6 +833,11 @@ typedef void * lv_font_user_data_t;
 #  define LV_SPRINTF_INCLUDE <stdio.h>
 #  define lv_snprintf     snprintf
 #  define lv_vsnprintf    vsnprintf
+#else   /*!LV_SPRINTF_CUSTOM*/
+#  ifndef CONFIG_LVGL_SPRINTF_DISABLE_FLOAT
+#    define CONFIG_LVGL_SPRINTF_DISABLE_FLOAT 0
+#  endif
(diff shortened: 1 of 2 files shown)


components/lvgl/lv_conf.h
```diff
@@ -833,6 +833,11 @@ typedef void * lv_font_user_data_t;
 #  define LV_SPRINTF_INCLUDE <stdio.h>
 #  define lv_snprintf     snprintf
 #  define lv_vsnprintf    vsnprintf
+#else   /*!LV_SPRINTF_CUSTOM*/
```
C47D, line 841: What do you think about doing something like this:
```c
#if defined CONFIG_LVGL_SPRINTF_DISABLE_FLOAT
#define LV_SPRINTF_DISABLE_FLOAT 0
#else
#define LV_SPRINTF_DISABLE_FLOAT 1
#endif
```

C47D:
Hi,

Didn't saw this configuration option until now, it was added with 0c4c6cf199362463c08fa5943eab38021bb753cb.

components/lvgl/lv_conf.h
an-erd: Yes, that's better. But we need to swap 0 and 1, it should read

```
#else   /*!LV_SPRINTF_CUSTOM*/
#  if defined CONFIG_LVGL_SPRINTF_DISABLE_FLOAT
#    define LV_SPRINTF_DISABLE_FLOAT 1
#  else
#    define LV_SPRINTF_DISABLE_FLOAT 0
#  endif
```

You want me to change the PR? (I'm not familiar with this suggestion function yet...)

PR #156, opened 2020-06-15
Fix: For ST7735 red and blue needs to be swapped

an-erd:
Colors are not correct for ST7735, red and blue swapped using ST7735_MADCTL


an-erd:
Need some more work first...

PR #158, opened 2020-06-15
Fix st7735 colors, inverted and set orientation cleanup; add default for LVGL_SPRINTF_CUSTOM

an-erd:


C47D:
I thought boolean symbols are set to n by default.

components/lvgl/Kconfig
```diff
@@ -572,6 +572,7 @@ menu "LVGL configuration"
     menu "Text Settings"
         config LVGL_SPRINTF_CUSTOM
             bool "Change the built-in (v)snprintf functions"
+            default false
```
embeddedt: I always thought the value had to be `y`, `n`, or `m`, and [`kconfig-language.txt`](https://www.kernel.org/doc/Documentation/kbuild/kconfig-language.txt) doesn't seem to mention `false`. For compatibility reasons, this might be worth looking into.

C47D: Yes, it needs to be `n`.

Also I think symbols defaulted to n are sometimes not defined, so maybe https://github.com/lvgl/lv_port_esp32/blob/master/components/lvgl/lv_conf.h#L837 should be
```c
#if defined LV_SPRINTF_CUSTOM
```

instead.

C47D: @an-erd could you please change the `false` to `n` and add the `#if defined LV_SPRINTF_CUSTOM` into a new commit?

Same on the `LVGL_SPRINTF_DISABLE_FLOAT` symbol, it defaults to `false`.

an-erd: Yes, you're right, it should read `n` for Kconfig.

In [https://github.com/lvgl/lv_port_esp32/blob/master/components/lvgl/lv_conf.h#L831](https://github.com/lvgl/lv_port_esp32/blob/master/components/lvgl/lv_conf.h#L831) there is already a `#if defined CONFIG_LVGL_SPRINTF_CUSTOM`, so `LV_SPRINTF_CUSTOM`should be either `1`or `0`, and thus the evaluation of `#if LV_SPRINTF_CUSTOM` should be fine, same as e.g. [here ](https://github.com/lvgl/lv_port_esp32/blob/master/components/lvgl/lv_conf.h#L817).

C47D:
Thanks for the fixes @an-erd , I will work on adding the LVGL widgets on the menuconfig.

PR #161, opened 2020-06-21
Add Debug/Text Settings to Kconfig and lv_conf.h

an-erd:


C47D:
Thanks @an-erd it looks good. Do you think there's something still left to be added on the menuconfig? Ican work on it.

PR #164, opened 2020-07-05
LVGL configuration from menuconfig

C47D:
@an-erd have contributed to add support to configure logging, debug and text settings to the Kconfig file used by menuconfig.

Some LVGL configuration options are not suitable for Kconfig, because they doesn't use Kconfig available types, we will leave those for later.

The end goal of this is add a Kconfig and a Kconfig based `lv_config.h` file to the main LVGL repo.


embeddedt:
In my humble opinion, needing to sync config options between `Kconfig` and `lv_conf_template.h` is going to result in extra work down the road. In the long run, I think it'd be better to stick with one central filetype (probably `Kconfig`), and generate the other one with a script.

C47D:
That is true @embeddedt, we are facing this very same issue when trying to work on this PR. I think it might be possible to create an Kconfig and `lv_config_kconfig_template.h` file from the `lv_config_template.h` file but it needs some cleanup on the comments so we can use them also on the menuconfig help.

embeddedt:
It might be better to do it the other way around (use `Kconfig` as the template and generate `lv_conf_template.h` from it). That might be simpler than parsing comments.

C47D:
Sorry, I don't follow, do you mean generating the current `lv_conf.h` file based on a `Kconfig` file?

C47D:
This PR is ready, I think the discussion for generating this files automatically can be done on LVGL, what do you think @embeddedt, @kisvegabor?

C47D:
Now that I think about it, the `Kconfig` and `lv_conf_kconfig_template` files are updated (if necessary) on each LVGL release. And I haven't seen the `lv_conf` file changing a lot. The initial work has been done here already.

components/lvgl/Kconfig
```diff
@@ -569,13 +569,418 @@ menu "LVGL configuration"
...
+        config LVGL_SPRINTF_CUSTOM
+        bool "Change the built-in (v)snprintf functions"
+        default n
+
         config LVGL_SPRINTF_DISABLE_FLOAT
             bool "Disable float in built-in (v)snprintf functions" if !LVGL_SPRINTF_CUSTOM 
             default n
     endmenu
 
+    menu "Objects usage"
```
kisvegabor: Can we replace "Object" and "OBJX" with "Widgets"?

C47D: Sure

C47D:
`LV_HOR_RES_MAX` and `LV_VER_RES_MAX` are on the display configuration menu.

> It'd be nice to make lv_conf_internal.h to work with Kconfig. The precedence in my opinion should be:
    Kconfig if present
    lv_conf.h
    Default values
What do you think?

I don't know how to implement this, the `lv_conf.h` on the lvgl directory is edited o it can take values assigned on the menuconfig, do you mean editing the lv_conf_internal work with it?

embeddedt:
I think we shouldn't try to mix the two systems. People should be able to use one or the other.

My suggestion was to use `Kconfig` to generate `lv_conf_template.h`, not `lv_conf.h`. That would reduce maintainer workload but still enable people to choose which config system they'd rather use.

components/lvgl/Kconfig
C47D: Done in 25ef9cc

C47D:
> My suggestion was to use Kconfig to generate lv_conf_template.h, not lv_conf.h. That would reduce maintainer workload but still enable people to choose which config system they'd rather use.

The problem I see with `Kconfig` is that it doesn't support all "C types", also the `lv_conf.h` file on this repo is different from the original LVGL `lv_conf.h`.

On this repo:
```
    #define LV_VER_RES_MAX          (CONFIG_LVGL_DISPLAY_WIDTH)
    #define LV_HOR_RES_MAX          (CONFIG_LVGL_DISPLAY_HEIGHT)
```

On the LVGL repo:
```
    #define LV_VER_RES_MAX          480
    #define LV_HOR_RES_MAX        320
```

That's why I was thinking on generate both `Kconfig` and `lv_conf_kconfig_template.h` from the `lv_conf.h`

kisvegabor:
I've just noticed that somehow my earlier message was not sent. It clarified my intention about adding Kconfig related thing in `lv_conf_internal.h`

So in `lv_conf_internal.h`
1. Check if Kconfig is in use. (E.g. by checking a config that is always present) 
2. If there is Kconfig do not include `lv_conf.h`
3. Add guards to options like this:
```c
#ifndef LV_FOO  /*LV_FOO*/
#  ifdef CONFIG_LVGL_FOO
#    define LV_FOO    CONFIG_LVGL_FOO
#  else
#    define LV_FOO    25
#  endif /*CONFIG_LVGL_FOO*/
#endif  /* ifndef LV_FOO*/

What do you think?
```

C47D:
It seems bloated to me, but if you want we can open an issue on LVGL repo with this so more people can give an opinion about it.

embeddedt:
`lv_conf_internal.h` is generated by a script, so having extra lines is not a huge problem.

C47D:
Ok, should we open an issue on LVGL for it?

kisvegabor:
> Ok, should we open an issue on LVGL for it?

Yes, please.

>  We should also create a Kconfig for LVGL v6.

Why? Do you think so many people are still using v6?

C47D:
Well, we need to create a branch for the v7 and v6 of LVGL in here.

kisvegabor:
IMO we can take a "snapshot" from the current structure (all components here) with submodules to v6 and call it v6 version of `lv_port_esp32`. 

Imagine we change something structural a year later. How long should we propagate the structural change back? In this regard, I think it fine to keep old releases as they are and focus on the current ones.

What do you think?

embeddedt:
I agree. I don't think structural changes should be propagated back to earlier releases, especially because it opens up the possibility of bugs. It's better to just keep an archive of the "last known working" version.

For example, I made a [`release/v5`](https://github.com/lvgl/lvgl/tree/release/v5) branch on the main lvgl repository to make it easier for users to redownload v5 for their existing projects if they need it, but that branch does not have any commits including and after the merge of `dev-6.0`, and won't receive any new commits.

C47D:
> IMO we can take a "snapshot" from the current structure (all components here) with submodules to v6 and call it v6 version of lv_port_esp32.

Cool, will do that over the weekend.

Can we merge this branch? There's a pending review.

kisvegabor:
Great!

We can merge it from my side. @embeddedt ?

embeddedt:
Looks good to me.

C47D:
Thanks, sorry for being so slow replying, I will try to keep up more frequently.

PR #167, opened 2020-07-20
Ttgo

limitz:
The TTGO uses some specific offsets as it uses the center area of the 135x240 LCD.
See https://github.com/Xinyuan-LilyGO/TTGO-T-Display


C47D:
Thanks for the PR @limitz, sorry for the late reply, I did some fixes on the Kconfig file.

PR #170, opened 2020-07-26
ili9488: RGB666 conversion

mcerveny:
5-bit to 6-bit color conversion (RGB565 to RGB666) should fill LSB with copy of MSB.

### DIFF
--- a/components/lvgl_esp32_drivers/lvgl_tft/ili9488.c
+++ b/components/lvgl_esp32_drivers/lvgl_tft/ili9488.c
@@ -128,11 +128,11 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
 
     for (uint32_t i = 0; i < size; i++) {
         LD = buffer_16bit[i].full;
-        mybuf[j] = (uint8_t) ((LD & 0xF800) >> 8);
+        mybuf[j] = (uint8_t) (((LD & 0xF800) >> 8) | ((LD & 0x8000) >> 13));
         j++;
         mybuf[j] = (uint8_t) ((LD & 0x07E0) >> 3);
         j++;
-        mybuf[j] = (uint8_t) ((LD & 0x001F) << 3);
+        mybuf[j] = (uint8_t) (((LD & 0x001F) << 3) | ((LD & 0x0010) >> 2));
         j++;
     }
 


C47D:
@mcerveny Sorry for the late reply, I though I already answered you, I'm not very familiar with color formats used by LVGL, maybe @embeddedt or @kisvegabor can help us a bit?

mcerveny:
The patch extends software conversion from RGB565 format to RGB666 format needed by ili948* in serial communication (MIPI-DBI Type C). RGB565 format is not supported in serial communication. The problem is with LSB in actual implementation - always zero. So for example, "white" is not never "white" in RGB666. Input 0xffff == RGB565 (31,63,31) -> RBG666 (62,63,62).  

rgb666convert

components/lvgl_esp32_drivers/lvgl_tft/ili9488.c
```diff
@@ -128,11 +128,11 @@ void ili9488_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * col
 
     for (uint32_t i = 0; i < size; i++) {
         LD = buffer_16bit[i].full;
-        mybuf[j] = (uint8_t) ((LD & 0xF800) >> 8);
+        mybuf[j] = (uint8_t) (((LD & 0xF800) >> 8) | ((LD & 0x8000) >> 13));
```
kisvegabor, line 131: It might be a little bit more efficien:
```c
 mybuf[j] = (uint8_t) (((LD & 0xF800) >> 8) | ((LD & 0x8000) ? 1 : 0));
```

mcerveny: I am not sure (not disassembled gcc optimized output for esp32) if "test/and" followed by "conditional **jump**" and "or with **constant**" is better then "test/and", followed by "rotate" and "or with register".
By the way the constant should be "4" not "1".

ili9488_serial_rgb666

kisvegabor: You are right, probably it really compiles to conditional jump so the difference is not so obvious. 

Anyway, the difference should be negligible, so let's leave your solution.

C47D:
It looks good to you as well @embeddedt?

C47D:
Will merge this, @kisvegabor maybe we should open an issue on LVGL to keep track of the test on 24bit support?

kisvegabor:
> maybe we should open an issue on LVGL to keep track of the test on 24bit support?

Good idea! Done here: https://github.com/lvgl/lvgl/issues/1722

PR #172, opened 2020-07-27
Initial support for ILI9481 TFT Controller

mcerveny:
ILI9481 is usually available with IPS with inverted colors.
ILI9481 SPI write speed is 10MHz by datasheet but working with 16MHz.


C47D:
Hi @mcerveny, everything looks right, but I don't have that display controller to test it. Is the orientation code working properly for all four options?

mcerveny:
> Is the orientation code working properly for all four options?
Yes, It works for me.

C47D:
Thanks for the contribution @mcerveny

PR #176, opened 2020-08-05
lvgl kconfig: Add some of the missing configuration options into the menuconfig interface

C47D:
What the title says :).

### DIFF
--- a/components/lvgl/Kconfig
+++ b/components/lvgl/Kconfig
@@ -774,7 +774,7 @@ menu "LVGL configuration"
 
     menu "Widgets"
         config LVGL_USE_OBJ_REALIGN
-            bool "Enable `lv_obj_realaign()` based on `lv_obj_align()` parameters."
+            bool "Enable `lv_obj_realign()` based on `lv_obj_align()` parameters."
             default y
         
         choice LVGL_EXT_CLICK_AREA
--- a/components/lvgl/Kconfig
+++ b/components/lvgl/Kconfig
@@ -1,5 +1,50 @@
 menu "LVGL configuration"
 
+    config LVGL_COLOR_SCREEN_TRANSP
+        bool "Enable screen transparency."
+        help
+            Useful for OSD or other overlapping GUIs.
+            Requires `LV_COLOR_DEPTH = 32` colors and the screen's style
+            should be modified: `style.body.opa = ...`.
+
+    config LVGL_ANTIALIAS
+        bool "Enable anti-aliasing (lines, and radiuses will be smoothed)."
+        default y
+
+    config LVGL_DISP_DEF_REFR_PERIOD
+        int "Default display refresh period (ms)."
+        default 30
+        help
+            Can be changed in the display driver (`lv_disp_drv_t`).
+
+    config LVGL_DPI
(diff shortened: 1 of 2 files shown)


C47D:
Most of the LVGL configuration options are now available on the menuconfig interface. Some options are not available because the Kconfig types are limited or those options are currently set-up when configuring the display used in the project.

I think the display dependent options need to be configured in the LVGL menu, this way LVGL can be used with other display drivers and not necessarily the available here, what do you think?

components/lvgl/Kconfig
```diff
@@ -712,7 +772,23 @@ menu "LVGL configuration"
             default n
     endmenu
 
-    menu "Widgets usage"
+    menu "Widgets"
+        config LVGL_USE_OBJ_REALIGN
+            bool "Enable `lv_obj_realaign()` based on `lv_obj_align()` parameters."
```
embeddedt: We may as well correct the spelling here at the same time.

```suggestion
            bool "Enable `lv_obj_realign()` based on `lv_obj_align()` parameters."
```

C47D: I will, this text is taken from the `lv_config_template.h`, we may fix it over there aswell.

kisvegabor: It seems already fixed in [lv_conf_templalte.h](https://github.com/lvgl/lvgl/blob/master/lv_conf_template.h#L538).

kisvegabor:
Looks good to me. 
Although there are some new defines in v7.3 as well: `LV_USE_OULINE/PATTERN/VALUE_STR` and `LV_MEMCPY_MEMSET_STD`.

As we already discussed turning lvgl to esp component would allow managing Kconfig by a script. 

@C47D you mentioned some types can't be added to Kconfig due to its limitations. Can you tell what were these?

C47D:
@kisvegabor The types we can not generate with Kconfig are the user defined ones, example the

```c
/* Type of coordinates. Should be `int16_t` (or `int32_t` for extreme cases) */
typedef int16_t lv_coord_t;
```

and

```
/*=====================
 *  Compiler settings
 *====================*/
/* Define a custom attribute to `lv_tick_inc` function */
#define LV_ATTRIBUTE_TICK_INC

/* Define a custom attribute to `lv_task_handler` function */
#define LV_ATTRIBUTE_TASK_HANDLER

/* Define a custom attribute to `lv_disp_flush_ready` function */
#define LV_ATTRIBUTE_FLUSH_READY

/* With size optimization (-Os) the compiler might not align data to
 * 4 or 8 byte boundary. This alignment will be explicitly applied where needed.
 * E.g. __attribute__((aligned(4))) */
#define LV_ATTRIBUTE_MEM_ALIGN

/* Attribute to mark large constant arrays for example
 * font's bitmaps */
#define LV_ATTRIBUTE_LARGE_CONST

/* Prefix performance critical functions to place them into a faster memory (e.g RAM)
 * Uses 15-20 kB extra memory */
#define LV_ATTRIBUTE_FAST_MEM
```

I will search what version of LVGL are we using here and add it at the top of the Kconfig file.

kisvegabor:
Thanks!

PR #179, opened 2020-08-07
IL3820 driver cleanup

C47D:
Preparing the IL3820 for portrait and landscape support, it currently crashes when configured as portrait.


C47D:
Got it working with the Hello World label of the demo project, will test it with more widgets and it definitely needs some code style improvements.

PR #181, opened 2020-08-09
子模块 添加缩放demo scrool

lukunn:


C47D:
Thanks for the contribution, we will add a configuration option on the menuconfig interface to configure the `lv_examples` demos so we will avoid hardcoding it into the main code.

PR #182, opened 2020-08-16
add comments to main

C47D:
Only some misc comments...

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -24,7 +24,9 @@
 #include "lvgl/lvgl.h"
 #include "lvgl_helpers.h"
 
+#ifndef CONFIG_LVGL_TFT_DISPLAY_MONOCHROME
 #include "lv_examples/src/lv_demo_widgets/lv_demo_widgets.h"
+#endif
 
 /*********************
  *      DEFINES
@@ -36,31 +38,26 @@
  *  STATIC PROTOTYPES
  **********************/
 static void lv_tick_task(void *arg);
-void guiTask(void *pvParameter);
-
+static void guiTask(void *pvParameter);
+static void create_demo_application(void);
 
 /**********************
  *   APPLICATION MAIN
  **********************/
 void app_main() {
     
-    //If you want to use a task to create the graphic, you NEED to create a Pinned task
-    //Otherwise there can be problem such as memory corruption and so on
+    /* If you want to use a task to create the graphic, you NEED to create a Pinned task
+     * Otherwise there can be problem such as memory corruption and so on.
+     * NOTE: When not using Wi-Fi nor Bluetooth you can pin the guiTask to core 0 */
     xTaskCreatePinnedToCore(guiTask, "gui", 4096*2, NULL, 0, NULL, 1);
 }
 
-static void lv_tick_task(void *arg) {
-    (void) arg;
-
-    lv_tick_inc(LV_TICK_PERIOD_MS);
-}
-
-//Creates a semaphore to handle concurrent call to lvgl stuff
-//If you wish to call *any* lvgl function from other threads/tasks
-//you should lock on the very same semaphore!
+/* Creates a semaphore to handle concurrent call to lvgl stuff
+ * If you wish to call *any* lvgl function from other threads/tasks
+ * you should lock on the very same semaphore! */
 SemaphoreHandle_t xGuiSemaphore;
 
-void guiTask(void *pvParameter) {
+static void guiTask(void *pvParameter) {
     
     (void) pvParameter;
     xGuiSemaphore = xSemaphoreCreateMutex();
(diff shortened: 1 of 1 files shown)


main/main.c
```diff
@@ -116,6 +122,27 @@ void guiTask(void *pvParameter) {
...
     ESP_ERROR_CHECK(esp_timer_start_periodic(periodic_timer, LV_TICK_PERIOD_MS * 1000));
 
+    /* Create the demo application */
+    create_demo_application();
+    
+    while (1) {
+        vTaskDelay(1);
+
+        /* Try to take the semaphore, call lvgl related function on success */
+        if (pdTRUE == xSemaphoreTake(xGuiSemaphore, (TickType_t) LV_TICK_PERIOD_MS)) {
```
embeddedt: I'm not sure if these lines are doing what you intend. It looks like you want to ensure that `lv_task_handler` gets called at least every `LV_TICK_PERIOD_MS` milliseconds. The current code looks like it will call `lv_task_handler` every 1ms as long as the semaphore is available.

I think this would make more sense:

```suggestion
        vTaskDelay(LV_TICK_PERIOD_MS);

        /* Try to take the semaphore, call lvgl related function on success */
        if (pdTRUE == xSemaphoreTake(xGuiSemaphore, portMAX_DELAY)) {
```

Since there's nothing else the task does if the semaphore is unavailable, it seems to me that it may as well just suspend until the semaphore becomes available.

C47D: You're right @embeddedt, fixed it with 0b8b819.

C47D:
@embeddedt I think you mentioned on the forum that the tick being called every 10ms is too long, do you think we should change it to 1ms instead?

embeddedt:
10ms should be fine... unless I've forgotten something. Do you know where I mentioned that?

C47D:
Can't find the particular post where I saw it, anyways, let's keep it at 10ms.

kisvegabor:
10 ms seems a little bit too much. The way a task with a 15 ms period will be called every 20 ms instead. The time measurement's accuracy is 10 ms as well. Therefore the performance monitor can measure only values like  0 ms, 10 ms, 20 ms, 30 ms, 40 ms, and get FPSs like 100 FPS, 50 FPS, 30 FPS, 25 FPS and nothing in between. 

So if possible I vote to have at least a 5 ms period for tick but 1 ms would be the best.

embeddedt:
Fair enough. For simple UIs that only use animations I don't know if the reduced frequency will be a major issue, as a 20ms refresh rate is still very close to 60fps. But since this is intended to be a ready-to-use project, we should try to have the performance monitor be as accurate as possible.

C47D:
Changed the tick to 1ms, can you review the change @kisvegabor?

main/main.c
```diff
@@ -116,6 +122,27 @@ void guiTask(void *pvParameter) {
     ESP_ERROR_CHECK(esp_timer_create(&periodic_timer_args, &periodic_timer));
     ESP_ERROR_CHECK(esp_timer_start_periodic(periodic_timer, LV_TICK_PERIOD_MS * 1000));
 
+    /* Create the demo application */
+    create_demo_application();
+    
+    while (1) {
+        vTaskDelay(LV_TICK_PERIOD_MS);
```
kisvegabor: Wouldn't be better to use `vTaskDelay(1)` to sleep for one heartbeat?
Now it's the same as `LV_TICK_PERIOD_MS == 1` but looks like someone though that `vTaskDelay` gets `ms` as parameter.

C47D: Does df5f3385eea49f050e416a77848dc12c15a793ac looks good?

kisvegabor: I think it will be solved as `vTaskDelay(0)` if tick period is 10 ms. According to [this](https://www.freertos.org/FreeRTOS_Support_Forum_Archive/December_2012/freertos_vTaskDelay_0_vs_vTaskDelay_1_6518023.html) `vTaskDelay(0)` is equivalent to `vTaskYield()`. Is this the desired behaviour?

C47D: The desired behavior is to delay for 1ms, that's why I used the macro to calculate the rtos ticks, `vTaskDelay` parameter is un ticks. Do you want me to change it to 1?

kisvegabor: Maybe I miss something but can it delay less than the tick period?

embeddedt: I think the delay is for the LVGL tick period not the RTOS tick period.

C47D: I'm confused now 😄, `vTaskDelay` parameter is in RTOS ticks by using the `pdMS_TO_TICKS` we convert `ms` to RTOS ticks.

kisvegabor: I see but if RTOS tick is 10 ms (ESP default) `pdMS_TO_TICKS(1)` will be `0`, isn't it?

PR #184, opened 2020-08-18
lv_examples: Add Kconfig file for menuconfig interface and add demo widgets configuration in it.

C47D:
Here´s the menuconfig interface so far, I'm thinking on letting the user choose which demo or example to run, what do you think?

imagen

### DIFF
--- a/components/lv_examples/Kconfig
+++ b/components/lv_examples/Kconfig
@@ -30,39 +30,6 @@ menu "LVGL examples configuration"
             - Montserrat 28 font.
             - Montserrat 32 font.
 
-        config LV_EXAMPLE_STYLE_1
-        bool "EXAMPLE STYLE 1: Using the background style properties."
-        
-        config LV_EXAMPLE_STYLE_2
-        bool "EXAMPLE STYLE 2: Using the border style properties."
-        
-        config LV_EXAMPLE_STYLE_3
-        bool "EXAMPLE STYLE 3: Using the outline style properties."
-        
-        config LV_EXAMPLE_STYLE_4
-        bool "EXAMPLE STYLE 4: Using the shadow style properties."
-        
-        config LV_EXAMPLE_STYLE_5
-        bool "EXAMPLE STYLE 5: Using the pattern style properties."
-        
-        config LV_EXAMPLE_STYLE_6
-        bool "EXAMPLE STYLE 6: Using the value style properties."
-        
-        config LV_EXAMPLE_STYLE_7
-        bool "EXAMPLE STYLE 7: Using the text style properties."
-        
-        config LV_EXAMPLE_STYLE_8
-        bool "EXAMPLE STYLE 8: Using the line style properties."
-        
-        config LV_EXAMPLE_STYLE_9
-        bool "EXAMPLE STYLE 9: Using the image style properties."
-        
-        config LV_EXAMPLE_STYLE_10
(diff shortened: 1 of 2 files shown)


kisvegabor:
I rarely use menuconfig so can't really decide if it's a good practice or not. ☹️ 

I suppose selecting the default demo still makes possible to add demo/code by hand. If so I think it won't harm anything besides the extra maintenance.

embeddedt:
I agree with @kisvegabor.

embeddedt:
Sorry; I hit the close button by accident.

C47D:
I agree, I thought this repo is to showcase LVGL in esp32, maybe not to be used as a project template, that's what the separated components are being created for, the user would include which they need. I will try to add the examples later this week, let's see how that plays out.

C47D:
@embeddedt @kisvegabor 

Current `lv_examples` menuconfig (not definitive), missing: _widget_ and _get started_ examples. It need some cosmetic changes, I don't personally like having all the available options always enabled. Will work on it this weekend.

imagen

What do you think?

kisvegabor:
@C47D 

Only the DEMOs can be enabled/disabled from the config. E.g. [lv_ex_style_1](https://github.com/lvgl/lv_examples/blob/master/src/lv_ex_style/lv_ex_style_1.c) has no `define` guards. It's because these examples are usually short and small so I think it's not worth the effort to enable/disable them one by one. So I think it's enough to have options already present in `lv_ex_conf.h`

C47D:
There's only two options on the `lv_ex_config` file, I will check it again tho.

C47D:
@kisvegabor I've removed the style examples from the menuconfig interface. It does look like this now:

imagen

kisvegabor:
In [lv_ex_conf_templ.h](https://github.com/lvgl/lv_examples/blob/a530172314d4221fcb1b4d2ba2344012961b9b3f/lv_ex_conf_templ.h) there are some more options. 

I think it works well if you select a demo to rum, it will enable it the demo itself as well. 
Do you plan to make examples selectable as well? I mean not to enable e.g. `lv_ex_btn_1()` just select it to run automatically.

C47D:
The config file you point out is newer, I would need to update `lv_examples`.

Can we merge this as first implementation and then update both `lv_examples` and `lvgl`. What do you think?

kisvegabor:
I meant e.g. for `LV_EXAMPLES_DEMO_BENCHMARK` you don't enable `LV_DEMO_BENCHMARK` in  [lv_ex_conf.h](https://github.com/lvgl/lv_port_esp32/pull/184/files#diff-1bbd24efc9c65645970725e6df604866).

C47D:
Because the `lv_ex_config` file doesn't have that option because it's old, we would need to update `lv_examples` to have it.

So I wanted to merge this, do a release of the repo, and then update both `lv_examples` and `lvgl`.

kisvegabor:
> So I wanted to merge this, do a release of the repo, and then update both lv_examples and lvgl.

Alright!

C47D:
Ok, I will merge this and another PR (#195), then do a release and then update both lvgl and lv_examples. Then update both config files and Kconfig files, if everything is working we could then consider adding the Kconfig file to the main repos.

C47D:
@kisvegabor is right, will open a new PR about this once `lv_examples` gets updated. Closing this.

Issue #183, opened 2020-08-16
Can TTGO Camera Plus development board be added?


HoGC:
I tried to configure, but no success
![56438efc5261fcb2f20356092eb63e3](https://user-images.githubusercontent.com/50402699/90333864-e7628400-dffb-11ea-8a3a-e6ca0fc2090c.png)
![c5264002da0353083682de80fd51752](https://user-images.githubusercontent.com/50402699/90333867-ef222880-dffb-11ea-897a-95ba79c9b433.png)
This is a ttgo project: https://github.com/lewisxhe/esp32-camera-screen

C47D:
Hi @HoGC, what do you mean with no success? The display didn't displayed anything? It displays garbage?

Can you paste here the output of `idf.py monitor`?

HoGC:
The display didn't displayed anything,  But the backlight is on.
 I can use this drive to display: https://github.com/Xinyuan-LilyGO/TTGO-T-Display
LOG
```
I (0) cpu_start: App cpu up.
I (335) heap_init: Initializing. RAM available for dynamic allocation:
I (342) heap_init: At 3FFAE6E0 len 00001920 (6 KiB): DRAM
I (348) heap_init: At 3FFC4950 len 0001B6B0 (109 KiB): DRAM
I (354) heap_init: At 3FFE0440 len 00003AE0 (14 KiB): D/IRAM
I (361) heap_init: At 3FFE4350 len 0001BCB0 (111 KiB): D/IRAM
I (367) heap_init: At 4008B9AC len 00014654 (81 KiB): IRAM
I (373) cpu_start: Pro cpu start user code
I (392) spi_flash: detected chip: generic
I (392) spi_flash: flash io: dio
I (393) cpu_start: Starting scheduler on PRO CPU.
I (0) cpu_start: Starting scheduler on APP CPU.
I (20) lvgl_helpers: Display hor size: 240, ver size: 240
I (20) lvgl_helpers: Display buffer size: 9600
I (20) lvgl_helpers: Initializing SPI master for display
I (30) lvgl_helpers: Configuring SPI host HSPI_HOST (1)
I (30) lvgl_helpers: MISO pin: -1, MOSI pin: 19, SCLK pin: 21
I (40) lvgl_helpers: Max transfer size: 19200 (bytes)
I (50) lvgl_helpers: Initializing SPI bus...
I (50) disp_spi: Adding SPI device
[...]

PR #186, opened 2020-08-21
Configure GPIOs pins as GPIO explicitly

C47D:
Some GPIOs (GPIO32 and GPIO33) need to be explicitly configured as gpios, as we don't know which pin are the users going to choose I think a good solution is to configure all the GPIOs.

Fixes part of #183, another PR will create the predefined display configuration for TTGO Camera Plus board.

### DIFF
--- a/components/lvgl_esp32_drivers/lvgl_tft/GC9A01.c
+++ b/components/lvgl_esp32_drivers/lvgl_tft/GC9A01.c
@@ -122,10 +122,13 @@ void GC9A01_init(void)
 #endif
 
 	//Initialize non-SPI GPIOs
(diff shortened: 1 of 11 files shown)


kisvegabor (commented):
Looks good!
Only the indentation is wrong in some places.  It might be worth considering some code formatting tools. lvgl uses [this](https://github.com/lvgl/lvgl/blob/master/scripts/code-format.sh).

kisvegabor:
> I've added astyle configuration file and script on #187

Great! I think we can merge it and run the formatted as you suggested.

C47D:
Thanks!

PR #189, opened 2020-08-27
Add TTGO Camera Plus predefined configuration

C47D:
Waiting for feedback from #183


C47D:
It was tested and it seems to be working properly.

PR #192, opened 2020-08-29
Initial GoodDisplay GDEW0154M09 / FitiPower JD79653A support

huming2207:
Hi @C47D ,

Here's a draft PR for GDEW0154M09, which is a new 1.54 inch e-paper panel from GoodDisplay, with FitiPower's JD79653A driver chip built-in.

Now I still have something need to be done first before you merge. That includes: 

- [x] Rotation support
- [x] General code cleanup
- [x] More testings
- [x] Configuration presets, e.g. SPI clock speed
- [x] Power-off and sleep 
- [x] Partial refresh - may have ghost marks

If you are convenient, please have a look for my code and provide some suggestions if possible. Thanks!

Regards,
Jackson


huming2207:
Here are some demo code and datasheet from GoodDisplay. Their official site is a bit slow (at least for me, in Australia), so I leave a backup here:

[G-GDEW0154M09-200810.zip](https://github.com/lvgl/lv_port_esp32/files/5144816/G-GDEW0154M09-200810.zip)

[GDEW0154M09-200709.pdf](https://github.com/lvgl/lv_port_esp32/files/5144818/GDEW0154M09-200709.pdf)

huming2207:
This panel is a 200x200 square panel, so it's a bit hard to say which is the landscape and which is the portrait mode. So here below I will mention it as degrees instead. At 0 deg it's the default way, shown as below.

So far according to the documentation, to turn 180 deg, just simply set both `SHL` and `UD` bit at register 0x00 to 0.

Here's when `SHL=1; UD=1` (0 deg)

7E7B90A2-153B-44A4-BB0C-FDF18E10502D

4D0A651D-2984-40C8-8397-1C8CC74A93AF

Here's when `SHL=0; UD=1` (content are mirrored)

483E81FC-41C9-47EE-AC4D-25EB6767BE34

Here's when `SHL=0; UD=0` (content looks good, turned 180 deg)

8CBF9B28-5C51-4B60-8B57-6CF8FDD401D4

But so far I have no idea how to turn the screen 90 deg or 270 deg. I have tried setting Gate/Source Start Setting register (0x65) but any non-default value will cause some weird shades at the back, while the main content is still visible. 

I will probably speak to the GD's customer support team and see if they can provide any details.

huming2207:
Hi @C47D ,

For the partial update, my idea for now is: 

1. Do partial refresh for a few times, say 10-ish times
2. Then get out of partial update mode, do one full refresh
3. After that, do another 10 times of partial refresh
4. After 10 times, do another full refresh
5. ...(another cycle)

I'm implementing this kind of cycles (or "state machine") like I mentioned above.

For previous IL3820's implementation, I remembered it was hard-coded to always refresh in partial mode. It may cause some ghost marks and/or damage the panel. In my opinion, this is not okay for production.

Regards,
Jackson

C47D:
Hi @huming2207,

That's a nice display 😄 , thanks for adding support for it!

> For previous IL3820's implementation, I remembered it was hard-coded to always refresh in partial mode. It may cause some ghost marks and/or damage the panel. In my opinion, this is not okay for production.

Yes, the IL3820 code is hackish, please correct me if I'm wrong, but the IL3820 code (flush function) is doing a full update of the display, and I remember reading on it's datasheet that partial update isn't available for that driver.

> For the partial update, my idea for now is:
> - Do partial refresh for a few times, say 10-ish times
> - Then get out of partial update mode, do one full refresh
> - After that, do another 10 times of partial refresh
> - After 10 times, do another full refresh
> - ...(another cycle)
> I'm implementing this kind of cycles (or "state machine") like I mentioned above.

Do you mean updating the display RAM without sending the "draw RAM content into the display" command? And then on the 10 update send the "draw RAM content into the display" command?

---

We are having some discussions about updating the [lv_drivers](https://github.com/lvgl/lv_drivers) repo with the drivers available in this repository, a re design of the drivers is being done so they become "platform independent", more info [here](https://forum.lvgl.io/t/merge-display-and-input-drivers-from-lv-port-esp32-to-lv-drivers/3021). This would allow users of LVGL use the variety of display and touch drivers we have in here.

When the new drivers architecture is ready, the `lvgl_esp32_drivers` directory will be moved into it's own repository, [here](https://github.com/lvgl/lvgl_esp32_drivers), after that we would need to port the drivers into the new architecture.

This will take some time, I only wanted to let you know the mid term goals on this repo :)

C47D:
> But so far I have no idea how to turn the screen 90 deg or 270 deg. I have tried setting Gate/Source Start Setting register (0x65) but any non-default value will cause some weird shades at the back, while the main content is still visible, but not rotated as I expected.
> I will probably speak to the GD's customer support team and see if they can provide any details.

I did that rotation in software, couldn't find any right configuration for the display registers. I hope to get some free time to update the IL3820 code in the near future.

huming2207:
Hi @C47D 

> > But so far I have no idea how to turn the screen 90 deg or 270 deg. I have tried setting Gate/Source Start Setting register (0x65) but any non-default value will cause some weird shades at the back, while the main content is still visible, but not rotated as I expected.
> > I will probably speak to the GD's customer support team and see if they can provide any details.
> 
> I did that rotation in software, couldn't find any right configuration for the display registers. I hope to get some free time to update the IL3820 code in the near future.


Yea agreed. Rotation is not a must-have feature anyway. I have disabled it for now. And again, I think LVGL itself should also provide its own software-based rotation support. This is very helpful for some panels with buggy controller or poor documentation.

> Do you mean updating the display RAM without sending the "draw RAM content into the display" command? And then on the 10 update send the "draw RAM content into the display" command?

For example, for some e-reader like Kindle or some Chinese ones, it usually does a few tens of partial refresh and then it will do a full refresh. This is to reduce the ghost mark/shadow.

> Yes, the IL3820 code is hackish, please correct me if I'm wrong, but the IL3820 code (flush function) is doing a full update of the display, and I remember reading on it's datasheet that partial update isn't available for that driver.

For IL3820, I think just leave it as-is for now until someone else wants to improve it. GoodDisplay will move to some new controller chip with DES panel later anyway. These old e-ink panels may be deprecated.

Regards,
Jackson

huming2207:
> We are having some discussions about updating the lv_drivers repo with the drivers available in this repository, a re design of the drivers is being done so they become "platform independent", more info here. This would allow users of LVGL use the variety of display and touch drivers we have in here.

For the platform issue, I think we should consider using some struct-based callbacks wrapper, so that we can make it to be in a kind of private function, public function and interfaces. Here's an example from Espressif: https://github.com/espressif/esp-idf/tree/c77c4ccf6c43ab09fd89e7c907bf5cf2a3499e3b/components/tcp_transport

Jackson

huming2207:
Hi again @C47D ,

So far I have done the partial refresh but it's not very perfect. The issues left are:

1. It seems like I have to leave it power-on all the time. Or otherwise, it will not get partial refreshes done properly. This power-off sequence works just fine when I only do full refreshes but no partial refreshes. GoodDisplay didn't document this, so I have no idea how to fix it anyway. I will have a chat with their customer support engineers later tomorrow.
2. The partial window must be the full screen as I have no idea how to extract out the updated field according to start and end coordinates from the framebuffer.
3. Partial refresh causes the whole screen to blink (for about 0.2 sec). This may be caused by the second issue mentioned above.

If you are happy to merge with these three issues remaining, then please proceed. I will also speak with GD's people next week. If I resolve these issues with them, I will get another PR submitted later when I am free. Probably next weekend.

Regards,
Jackson

huming2207:
Also another strange issue. It seems like if SPI clock is faster than 1MHz, the partial refresh will stop working.

It supposed to handle 10MHz for write operations but I have no idea why it doesn't work...Did I misinterpret something?

image

image

huming2207:
Hi @C47D 

I've contacted GD customer support team but their engineer was absent today. But anyway I've just found a workaround and fixed the issues.

Now the minor issues left are:

1. If it keeps refreshing it for 15+ rounds in less than 50 seconds, then it may have some ghost marks. Even a full refresh can't get rid of it. It must take a short break (say another 1 minutes or so) to recover.
2. Partial refresh is now actually refreshing the whole 200x200 pixels, instead of the actual fragment that it needs to be refreshed. But still, this is better than full refreshes, as it will be less blinky (~0.35s vs ~0.93s at 15-20 deg room temperature). 

The second issue may be related to the first issue. But since these non-DES e-paper display shouldn't refresh that often anyway (otherwise either battery or the panel will die very soon), it won't be too problematic for the general use cases.

I think I have done my work for now, please consider merging this PR if you are happy to do so. If not, please provide more suggestions. Thanks.

Regards,
Jackson

huming2207:
Hi @C47D ,

One more thing I forgot to mention. I've also got a GDEW0154M10 sample from GoodDisplay as well, which is another 1.54 inch panel but using UC8151D + 152x152 DES technology. GD will move to DES technology soon in the following months, as it is cheaper with better performance. This GDEW0154M10 is one of their experimental products currently available. 

I will get another PR done apart from this one. But in case of code conflict or extra confusion, could you please consider review and approve this PR soon? Thanks!

Jackson

C47D:
Hi @huming2207,

Sorry for the late reply, I was out this weekend.

> I think I have done my work for now, please consider merging this PR if you are happy to do so. If not, please provide more suggestions. Thanks.

I will check the PR later today, can't provide suggestions right now because I haven't read the data sheet of the part just yet. I agree you can add further features later.

The new technology looks interesting, I will try to look at it when I have some time and hopefully access to the hardware.

Thanks for your continuous help :)

Carlos

huming2207:
Hi @C47D ,

Thanks for your reply. GoodDisplay's customer service engineers just came back to me with some of their suggestions. I may do another one or two commits later after they finish reviewing my code.

> The new technology looks interesting, I will try to look at it when I have some time and hopefully access to the hardware.

I think the GDEW0154M10 is still in a pre-mass production phrase anyway. You may not get the panels from AliExpress or somewhere directly. But I think you can try to contact GD directly and see if they can sell you one or two.

In regards to your previous thoughts about the all-in-one driver, I have some more suggestions. Currently, all the `flush`/`rounder`/`set_px` callbacks are all left into the `lv_tft/disp_driver.c`. In my opinion, is not a good practice. The proper way may be passing the pointer of `lv_disp_drv_t`. For example:

- In each panel driver source file, say my `jd79653a.c`, has a init function 
   - like `void jd79653a_init(lv_disp_drv_t *drv) ...` 
- In my driver source file, it also has some static functions, that includes the `flush`/`rounder`/`set_px` callbacks.
- Inside of that init function, it assigns the `flush`/`rounder`/`set_px` callbacks to that `drv` pointer. 
   - Maybe something like `drv->flush_cb = jd79653_flush;`
- Same principles also apply for SPI/I2C drivers. Just use a struct of function pointers to deal with it. The user can choose to assign different types of SPI/I2C drivers depends on which chip they use.

Regards,
Jackson

C47D:
> In regards to your previous thoughts about the all-in-one driver, I have some more suggestions. Currently, all the flush/rounder/set_px callbacks are all left into the lv_tft/disp_driver.c. In my opinion, is not a good practice. What if later you have a few hundreds of panels in your driver?
> 
> The proper way may be passing the pointer of lv_disp_drv_t. For example:
> 
>     In each panel driver source file, say my jd79653a.c, has a init function
>         like void jd79653a_init(lv_disp_drv_t *drv) ...
>     In my driver source file, it also has some static functions, that includes the flush/rounder/set_px callbacks.
>     Inside of that init function, it assigns the flush/rounder/set_px callbacks to that drv pointer.
>         Maybe something like drv->flush_cb = jd79653_flush;
>     Same principles also apply for SPI/I2C drivers. Just use a struct of function pointers to deal with it. The user can choose to assign different types of SPI/I2C drivers depends on which chip they use.

I was also thinking on user callbacks, I need to take some time and draft it.

Let me know when the PR is ready for review :)

Regards
Carlos

huming2207:
Hi @C47D 

Yep it's ready. 

GD's people said the ghost mark issue is inevitable anyway, due to the nature of e-paper materials. So the only way to prevent that is to 1. do a bit more full refreshes (now is a full refresh after every 5 partials) 2. reduce the duty cycle (don't use it too often, like updating 20 times every one minute).

Regards,
Jackson

C47D:
Thanks @huming2207, I will check it as soon as possible.

Regards
Carlos

huming2207:
Hi @C47D ,

Yep no problem. Please proceed. Thanks.

The only thing left is the 90/270 deg rotation. But I think LVGL should provide that feature instead of letting the screen to do it anyway.  So I left it undone for now.

Jackson

C47D:
Hi @huming2207,

After taking a look at the PR it seems correct. Thanks for the contribution!

Regards,
Carlos

PR #195, opened 2020-09-02
Initial UC8151D / GoodDisplay GDEW0154M10 DES e-paper panel support

huming2207:
Hi @C47D ,

Here's the initial support for GDEW0154M10 or UC8151D. This is the world's first e-paper panel with [DES technology](http://www.e-paper-display.com/download_detail/downloadsId%3d942.html), with 152x152 resolution and in 1.54-inch size.

9D326422-A9C2-4AFD-AE0F-ECF5E6B4B714

Please note that:
1. I'm porting this code with a probable pre-production sample panel (and probably this is the first e-paper panel in Australia that with DES technology...).
2. This panel does not support partial refresh, at least for now. The full refresh is about 7.2 seconds at 20 deg room temperature.
3. I guess UC8151D might be an old revision of JD79653A, as you will see the registers are 90% the same.
4. Since they are very similar, I think this panel should support 180deg rotation as well. But I need to get it implemented and tested first. I will do that next.
5. So far I think I can't do too many optimisations as there is no much documentation available.

Regards,
Jackson


huming2207:
Hi again @C47D ,

Here's the 180 deg rotation:

19F42BFC-ACCB-48CF-84B1-9DCDB8677E93

I think I've done all the stuff I can for now. Please consider review and approve this PR. Thanks.

Regards,
Jackson

C47D:
Nice, you're fast @huming2207, could you add `if(ESP_PLATFORM)` at the beginning of the `CMakeLists.txt` and `end()` at the end? See bec7974 for reference.

huming2207:
Hi @C47D ,

Sure. Here you go.

Issue #201, opened 2020-09-24
Adding support for Dual and Quad SPI modes


davidjade:
I've been working on extending the current driver to support both Dual and Quad SPI and I've got it working with the FT81x but there are some serious caveats - the main one being that reading more than 1 byte of data at a time does not appear to working in HALF_DUPLEX mode using a mix of address and dummy bits as required by the FT81x. At this point I've spent a lot of time trying to understand what is going on and from looking at logic analyzer traces and diving deep into the ESP32 hal and spi_master, I am at a loss to what it happening but I can work around it. I've raise the issue with Espressif and I'm hoping someone has some answers because the driver seems broken in this "half-duplex read with address and dummy bits" scenario that the FT81x requires. Here's all the gory details I sent to Espressif: https://esp32.com/viewtopic.php?f=14&t=17436&sid=09557265fac47f98e92f722e94c82774

The good new is that writing data seems to always work reliably and I've got the FT81x humming along at 32Mhz in Dual mode on a breadboard no problem. Quad mode also works but I'm finding that line noise on a breadboard is a problem so I'm using Dual mostly.

To make it work for reading though, I just read single bytes at time and re-constitute the uint16_t or uint32_t that are needed with the FT81x. It works and isn't a huge penalty since I rarely read more the 4 bytes at a time. I can also isolate this work around to the FT81x driver itself to keep the mess out of the disp_spi code with one exception - there is a weirdness about specifying dummy bits (which I suspect is a documentation error - they appear to be dummy clocks, not dummy bits when I look at the logic traces).

So the question is, given that there are some issues that may be found when reading data in Dual and Quad modes, should my changes make it back to the lv_port_esp32 at this point? On the one hand only the FT81x is special in that there is no separate touch driver and all the other display types with separate touch drivers probably won't be affected (as I have not touched them). But I also don't have other displays that support Dual or Quad modes to test on, nor do I have the other displays that are already supported to test that I haven't broken anything.

C47D:
Hi,

Thanks for the link, I find very useful to read others experiences.

> So the question is, given that there are some issues that may be found when reading data in Dual and Quad modes, should my changes make it back to the lv_port_esp32 at this point?

Actually we are trying to contribute the display and touch controllers back to lv_drivers. There's a post on the forum about it, [here](https://forum.lvgl.io/t/merge-display-and-input-drivers-from-lv-port-esp32-to-lv-drivers/3021).

[...]

PR #202, opened 2020-09-27
Added support for Dual and Quad SPI modes and updating FT81x driver to use it

davidjade:
This resolves issue #201 by updating the base SPI display driver code to support Dual and Quad SPI modes as well as updating the FT81x drivers to use those modes. New kconfig options where added to control SPI modes, pins, etc... These changes should have minimal to no impact on the other existing drivers as using DIO/QIO modes are opt-in and the code path for not using it remains the same. 

Note that I also normalized to spaces in kconfig as using mixed tabs can cause ESP-IDF to freak out in some cases. This might make the kconfig look like it has changed more than it really has.

This update doubles and quadruples the refresh rate for the FT81x when using Dual or Quad SPI modes.


C47D:
Hi,

Sorry for the late reply. I won't be able to review this until next weekend, hope you can wait for it.

davidjade:
I did a careful review and fixed up the other drivers but since they are not using the new DIO/QIO features, they shouldn't even notice. The extra parameter (dummy bits) would only change behavior if they also passed in the new flag to activate them.

PR #205, opened 2020-10-01
Switch to using a pool of DMA transactions for reduced latency

### LINKED ISSUE #203: Adding support for multiple queued DMA transactions
After adding the DIO/QIO SPI modes I have done some further experiments with enabling a queue of SPI in-flight transactions and the results so far are very promising. By using a queue of in-flight DMA transaction structures instead of having just one static in-flight transaction, the Esp32 DMA controller can be sent multiple SPI transactions to process instead of the current code, where it always waits for the previous one to be finished before sending the next. In the current code this is a bottleneck that slows down redraw speeds and can also lead to a jerky display especially while animating. This bottleneck also likely prevents LVGL from using both update buffers effectively as well (since it can block the update thread).

Where this really pays off though, is for drivers that have to send data line by line (because the hardware doesn't support 7an arbitrary rectangle). The FT81x is one such devices and unless the rectangle is full display width, it has to break up repaint region rectangle into individual SPI DMA requests for each line of the area in a fast loop. Enabling 20, 50 or even 200 SPI transactions to be in-flight at once has a very, very noticeable effect of the speed of the display. The difference is night and day in refresh rates especially when using animated controls, etc... It's the first time I have seen the FT81x animate while being consistently free of animation frame skipping. Here's a quick video of the before/after: https://youtu.be/HxO02oeiakw

Since I laid some of this groundwork a few months back this isn't looking to be too hard to add this feature. All the changes will be in disp_spi.c. This also means that if there are other drivers that also loop over or burst sending SPI data (which results in multiple DMA transactions), they will automatically benefit from queued DMA transactions. For instance, the FT81x driver needed no changes at all. I don't expect any existing drivers to break (or even know) either. 

At this point I have this working and I should have it ready to submit in a day or so, once I put it through some more testing, cleanup, etc... It builds on the Dual/Quad pull request I submitted earlier.

davidjade:
This update switches SPI displays to used a pool of spi_transaction_ext_t structures for DMA transactions and resolves #203. It reduces DMA transaction latency and thus increases overall transfer throughput. This is completely transparent to all existing drivers but for those that burst multiple DMA requests (such as the FT81x, il3820, jd79653a, uc8151d, etc...) it can have a really big impact on overall display responsiveness and animation speeds. It also better supports LVGL's dual buffer mechanism by making the SPI thread wait less.

This is enabled for all existing display drivers but the old behavior can be re-enabled by setting SPI_TRANSACTION_POOL_SIZE to 1. That effectively reverts to the old behavior of sending 1 transactions and a time and waiting for it to complete before sending more. I don't think this is necessary though as this change only really reduces the latency between DMA requests when bursting data and existing drivers should not care (or notice). 

I also have some more detailed notes on the design of the transaction pool but I'm not sure where to put them in this project. It's about 1/2 a page of so in length and just describes some design decisions, nuances, etc...

### DIFF
--- a/components/lvgl_esp32_drivers/lvgl_helpers.c
+++ b/components/lvgl_esp32_drivers/lvgl_helpers.c
@@ -207,8 +207,9 @@ bool lvgl_spi_driver_init(int host,
     };
 
     ESP_LOGI(TAG, "Configuring SPI host %s (%d)", spi_names[host], host);
-    ESP_LOGI(TAG, "MISO pin: %d, MOSI pin: %d, SCLK pin: %d",
-        miso_pin, mosi_pin, sclk_pin);
+    ESP_LOGI(TAG, "MISO pin: %d, MOSI pin: %d, SCLK pin: %d, IO2/WP pin: %d, IO3/HD pin: %d",
+        miso_pin, mosi_pin, sclk_pin, quadwp_pin, quadhd_pin);
+
     ESP_LOGI(TAG, "Max transfer size: %d (bytes)", max_transfer_sz);
 
     spi_bus_config_t buscfg = {
--- a/components/lvgl_esp32_drivers/lvgl_tft/disp_spi.c
+++ b/components/lvgl_esp32_drivers/lvgl_tft/disp_spi.c
@@ -27,6 +27,40 @@
 #include "../lvgl_helpers.h"
 #include "../lvgl_spi_conf.h"
 
+/******************************************************************************
+ * Notes about DMA spi_transaction_ext_t structure pooling
+ * 
+ * An xQueue is used to hold a pool of reusable SPI spi_transaction_ext_t 
+ * structures that get used for all DMA SPI transactions. While an xQueue may 
+ * seem like overkill it is an already built-in RTOS feature that comes at 
+ * little cost. xQueues are also ISR safe if it ever becomes necessary to 
+ * access the pool in the ISR callback.
+ * 
+ * When a DMA request is sent, a transaction structure is removed from the 
+ * pool, filled out, and passed off to the esp32 SPI driver. Later, when 
+ * servicing pending SPI transaction results, the transaction structure is 
+ * recycled back into the pool for later reuse. This matches the DMA SPI 
+ * transaction life cycle requirements of the esp32 SPI driver.
+ * 
+ * When polling or synchronously sending SPI requests, and as required by the 
+ * esp32 SPI driver, all pending DMA transactions are first serviced. Then the 
+ * polling SPI request takes place. 
+ * 
+ * When sending an asynchronous DMA SPI request, if the pool is empty, some 
+ * small percentage of pending transactions are first serviced before sending 
+ * any new DMA SPI transactions. Not too many and not too few as this balance 
+ * controls DMA transaction latency.
+ * 
+ * It is therefore not the design that all pending transactions must be 
+ * serviced and placed back into the pool with DMA SPI requests - that 
+ * will happen eventually. The pool just needs to contain enough to float some 
+ * number of in-flight SPI requests to speed up the overall DMA SPI data rate 
+ * and reduce transaction latency. If however a display driver uses some 
+ * polling SPI requests or calls disp_wait_for_pending_transactions() directly,
+ * the pool will reach the full state more often and speed up DMA queuing.
+ * 
+ *****************************************************************************/
+
 /*********************
  *      DEFINES
  *********************/
--- a/components/lvgl_esp32_drivers/lvgl_tft/disp_spi.c
+++ b/components/lvgl_esp32_drivers/lvgl_tft/disp_spi.c
@@ -30,6 +30,15 @@
 /*********************
  *      DEFINES
  *********************/
+#define SPI_TRANSACTION_POOL_SIZE 50	/* maximum number of DMA transactions simultaneously in-flight */
+
+/* DMA Transactions to reserve before queueing additional DMA transactions. A 1/10th seems to be a good balance. Too many (or all) and it will increase latency. */
+#define SPI_TRANSACTION_POOL_RESERVE_PERCENTAGE 10
+#if SPI_TRANSACTION_POOL_SIZE >= SPI_TRANSACTION_POOL_RESERVE_PERCENTAGE
+#define SPI_TRANSACTION_POOL_RESERVE (SPI_TRANSACTION_POOL_SIZE / SPI_TRANSACTION_POOL_RESERVE_PERCENTAGE)	
+#else
+#define SPI_TRANSACTION_POOL_RESERVE 1	/* defines minimum size */
+#endif
 
 /**********************
  *      TYPEDEFS
@@ -45,7 +54,7 @@ static void IRAM_ATTR spi_ready (spi_transaction_t *trans);
  **********************/
 static spi_host_device_t spi_host;
 static spi_device_handle_t spi;
-static volatile uint8_t spi_pending_trans = 0;
+static QueueHandle_t TransactionPool = NULL;
 static transaction_cb_t chained_post_cb;
 
 /**********************
@@ -80,7 +89,7 @@ void disp_spi_add_device_with_speed(spi_host_device_t host, int clock_speed_hz)
         .mode = SPI_TFT_SPI_MODE,
         .spics_io_num=DISP_SPI_CS,              // CS pin
         .input_delay_ns=DISP_SPI_INPUT_DELAY_NS,
-        .queue_size=1,
+        .queue_size=SPI_TRANSACTION_POOL_SIZE,
         .pre_cb=NULL,
         .post_cb=NULL,
 #if defined(DISP_SPI_HALF_DUPLEX)
@@ -95,6 +104,19 @@ void disp_spi_add_device_with_speed(spi_host_device_t host, int clock_speed_hz)
     };
 
     disp_spi_add_device_config(host, &devcfg);
+
+	/* create the transaction pool and fill it with ptrs to spi_transaction_ext_t to reuse */
+	if(TransactionPool == NULL) {
+		TransactionPool = xQueueCreate(SPI_TRANSACTION_POOL_SIZE, sizeof(spi_transaction_ext_t*));
+		assert(TransactionPool != NULL);
+		for (size_t i = 0; i < SPI_TRANSACTION_POOL_SIZE; i++)
+		{
+			spi_transaction_ext_t* pTransaction = (spi_transaction_ext_t*)heap_caps_malloc(sizeof(spi_transaction_ext_t), MALLOC_CAP_DMA);
+			assert(pTransaction != NULL);
+			memset(pTransaction, 0, sizeof(spi_transaction_ext_t));
+			xQueueSend(TransactionPool, &pTransaction, portMAX_DELAY);
+		}
+	}
 }
 
 void disp_spi_change_device_speed(int clock_speed_hz)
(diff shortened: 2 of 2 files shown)


davidjade:
Here are those extra design notes

```c
/******************************************************************************
 * Notes about DMA spi_transaction_ext_t structure pooling
 * 
 * An xQueue is used to hold a pool of reusable SPI spi_transaction_ext_t 
 * structures that get used for all DMA SPI transactions. While an xQueue may 
 * seem like overkill it is an already built-in RTOS feature that comes at 
 * little cost. xQueues are also ISR safe if it ever becomes necessary to 
 * access the pool in the ISR callback.
 * 
 * When a DMA request is sent, a transaction structure is removed from the 
 * pool, filled out, and passed off to the esp32 SPI driver. Later, when 
 * servicing pending SPI transaction results, the transaction structure is 
 * recycled back into the pool for later reuse. This matches the DMA SPI 
 * transaction life cycle requirements of the esp32 SPI driver.
 * 
 * When polling or synchronously sending SPI requests, and as required by the 
 * esp32 SPI driver, all pending DMA transactions are first serviced. Then the 
 * polling SPI request takes place. 
 * 
 * When sending an asynchronous DMA SPI request, if the pool is empty, some 
 * small percentage of pending transactions are first serviced before sending 
 * any new DMA SPI transactions. Not too many and not too few as this balance 
 * controls DMA transaction latency.
 * 
 * It is therefore not the design that all pending transactions must be 
 * serviced and placed back into the pool with DMA SPI requests - that 
 * will happen eventually. The pool just needs to contain enough to float some 
 * number of in-flight SPI requests to speed up the overall DMA SPI data rate 
 * and reduce transaction latency. If however a display driver uses some 
 * polling SPI requests or calls disp_wait_for_pending_transactions() directly,
 * the pool will reach the full state more often and speed up DMA queuing.
 * 
 *****************************************************************************/ 
```

embeddedt:
How about adding the notes to `disp_spi.c`?

davidjade:
Yes I can do that. Just wanted coding guideline clarification before I added a huge block of comments.

C47D:
Hey @davidjade,

Thanks for the PR, is it possible for you to enable `LV_USE_PERF_MONITOR` in the configuration file so we can have a comparison between the old code and this one? I saw your video and the change is impressive.

As commented on #203, I will test this on all the displays I have, IL3820 in particular, that driver is a bit hacky so it have a lot of room for improvement.

davidjade:
Hi @C47D 

I ran some tests with LV_USE_PERF_MONITOR enabled and then selected the Visual tab, and scrolled to shows the 3 animated dial graphs on the display. Here are the results I get with the FT81x:

* With the old single transaction mode: flips between either 3 -or- 33 FPS and 17% CPU (but the display is skipping many frames)
* With the new pooling transaction mode: stable 14-16 FPS and 5-6% CPU but occasionally jumps to higher CPU loads in the 25-25% range for short periods (and no frames appear to be skipped)

In single mode, I think LVGL realizes it must throw out frames to stay on time for animations and that artificially leads to higher frame rates, hence by the numbers single mode looks better for FPS but poorer visually than pooling mode looks.

On the first screen, which is largely idle, the numbers report:
* single model: 11 or 33 FPS and 10-32% CPU 
* pooling : solid 33 FPS but only 5-6% CPU.

So I think what we're seeing is that the pooling mode is more efficient in CPU while having more consistent frame rates. In heavier animated screens, the pooling mode can push a steady 15-15 FPS consistently but of course that sometimes comes with an increased CPU use as well since more frames are being shown.

Overall the FPS consistency of pooling mode and the ability to more effectively use the CPU for heavy animations seems like a win.

Is that the sort of info you were looking for?

C47D:
I tried to test this PR but I can't get my display (ILI9341) to work, the display is working because I'm using it with other prototype. Will try to get it working again and then test this PR.

davidjade:
Hmm. Doesn't the WROVER 4.1 Kit have an ILI9341? I just dug one out (forgot I had it). it seems to work for me. I just used the setup for the predefined WROVER 4.1 Kit in menuconfig.

Maybe in your setup it doesn't like the faster rates. You could try setting SPI_TRANSACTION_POOL_SIZE to 1 (in disp_spi.c) to revert to the old behavior. It's slightly different code to count the transactions to wait on but if you do revert via the 1 settings, it will do exactly what it did before, which is always wait to clear the previous transaction before sending the next.

C47D:
Hi,

Yes, the Wrover does include an ILI9341 display, I think the problem is my flying wires. Can you attach the generated `sdkconfig.h` file? And paste the `idf.py monitor` output if you don't mind.

I will try your suggestion as well.

davidjade:
Hi @C47D 

Here's the serial output and sdkconfig.h file

[sdkconfig.h.txt](https://github.com/lvgl/lv_port_esp32/files/5322044/sdkconfig.h.txt)

```
rst:0x1 (POWERON_RESET),boot:0x3e (SPI_FAST_FLASH_BOOT)
flash read err, 1000
ets_main.c 371
ets Jun  8 2016 00:22:57

rst:0x10 (RTCWDT_RTC_RESET),boot:0x3e (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3fff0030,len:4
load:0x3fff0034,len:7180
load:0x40078000,len:13856
load:0x40080400,len:4052
entry 0x40080688
I (29) boot: ESP-IDF 3.40100.200827 2nd stage bootloader
I (29) boot: compile time 00:52:57
I (30) boot: chip revision: 1
I (33) boot_comm: chip revision: 1, min. bootloader chip revision: 0
I (42) boot.esp32: SPI Speed      : 40MHz
I (45) boot.esp32:        SPI Mode       : DIO
I (49) boot.esp32: SPI Flash Size : 4MB
I (54) boot: Enabling RNG early entropy source...
I (59) boot: Partition Table:
I (63) boot: ## Label            Usage          Type ST Offset   Length
I (70) boot:  0 nvs              WiFi data        01 02 00009000 00006000
I (77) boot:  1 phy_init         RF data          01 01 0000f000 00001000
I (85) boot:  2 factory          factory app      00 00 00010000 00300000
I (92) boot:  3 coredump         Unknown data     01 03 00310000 00010000
I (100) boot: End of partition table
I (104) boot_comm: chip revision: 1, min. application chip revision: 0
I (111) esp_image: segment 0: paddr=0x00010020 vaddr=0x3f400020 size=0x11124 ( 69924) map
I (151) esp_image: segment 1: pad       dr=0x0002114c vaddr=0x3ffb0000 size=0x02284 (  8836) load
I (155) esp_image: segment 2: paddr=0x000233d8 vaddr=0x40080000 size=0x00404 (  1028) load
I (158) esp_image: segment 3: paddr=0x000237e4 vaddr=0x40080404 size=0x0c834 ( 51252) load
I (191) esp_image: segment 4: paddr=0x00030020 vaddr=0x400d0020 size=0x4c60c (312844) map
I (327) esp_image: segment 5: paddr=0x0007c634 vaddr=0x4008cc38 size=0x027d8 ( 10200) load
I (341) boot: Loaded app from partition at offset 0x1       0000
I (341) boot: Disabling RNG early entropy source...
I (342) psram: This chip is ESP32-D0WD
I (347) spiram: Found 32MBit SPI RAM device
I (351) spiram: SPI RAM mode: flash 40m sram 40m
I (356) spiram: PSRAM initialized, cache is in low/high (2-core) mode.
I (363) cpu_start: Pro cpu up.
I (367) cpu_start: Application information:
I (372) cpu_start: Project name:     lvgl-demo
I (377) cpu_start: App version:      v1.0-20-gee274ff-dirty
I (383) cpu_start: Compile time:     Oct  3 2020 01:04:21
I (389) cpu_start: ELF file SHA256:  ad9ab0104a360135...
I (395) cpu_start: ESP-IDF:          3.40100.200827
I (401) cpu_start: Starting app cpu, entry point is 0x40081ef0
I (0) cpu_start: App cpu up.
I (1303) spiram: SPI SRAM memory test OK
I (1303) heap_init: Initializing. RAM available for dynamic allocation:
I (1304) heap_init: At 3FFAE6E0 len 00001920 (6 KiB): DRAM
I (1310) heap_init: At 3FFCF358 len 00010CA8 (67 KiB): DRAM
I (1316) heap_init: At 3FFE0440 len 00003AE0 (14 KiB): D/IRAM
I (1322) heap_init: At 3FFE4350 len 0001BCB0 (111 KiB): D/IRAM
I (1329) heap_init: At 4008F410 len 00010BF0 (66 KiB): IRAM
I (1335) cpu_start: Pro cpu start user code
I (1340) spiram: Adding pool        of 4096K of external SPI memory to heap allocator
I (1361) spi_flash: detected chip: gd
I (1362) spi_flash: flash io: dio
I (1362) cpu_start: Starting scheduler on PRO CPU.
I (0) cpu_start: Starting scheduler on APP CPU.
I (1369) spiram: Reserving pool of 32K of internal memory for DMA/internal allocations
I (1389) lvgl_helpers: Display hor size: 320, ver size: 240
I (1389) lvgl_helpers: Display buffer size: 20480
I (1399) lvgl_helpers: Initializing SPI master for display
I (1399) lvgl_helpers: Configuring SPI host VSPI_HOST (2)
I (1399) lvgl_helpers: MISO pin: -1, MOSI pin: 23, SCLK pin: 19, IO2/WP pin: -1, IO3/HD pin: -1
I (1419) lvgl_helpers: Max transfer size: 40960 (bytes)
I (1419) lvgl_helpers: Initializing SPI bus...
I (1419) disp_spi: Adding SPI device
I (1429) disp_spi: Clock speed: 40000000Hz, mode: 0, CS pin: 22
I (1639) ILI9341: Initialization.
I (1839) ILI9341: Enabling backlight.
I (1839) ILI9341: Display orientation: LANDSCAPE
I (1839) ILI9341: 0x36 command value: 0x28
```

C47D:
Hi @davidjade,

Thanks for the information, everything looks the same on my side, I merged a commit two days ago, does your main (88df263ea3d50a1569503fff3d0028897df981e4) have those changes?

davidjade:
Ok, git is not my strong point but I think I managed to get my fork with my SPI changes rebased onto the current public master. let me know if you need anything else on my end.

C47D:
Don't worry, neigther is mine. I've tested it with my 240x240 ST7795 based display and I don't see any regressions in performance.

When you pointed out the IL3820 driver could be improved with this, did you meant "free" improvements (with the current driver as is) or did you meant there's new improvement opportunities thanks to this PR?

In any case, I will merge this PR tomorrow after testing with an ILI9488 display I have in hand, I'm really curious why the ILI9341 based display didn't worked and it's working fine in another prototype.

Thanks for the help @davidjade.

davidjade:
I meant "free" improvements. I think a few existing drivers look like they break up each LVGL update and write it out line by line or block by block, etc... in a loop - much like the FT81x has to do in some cases. So they should see less latency between these individual SPI transfers now as they break an update into chunks, which should increase perceived refresh speeds.

There may be other opportunities now as well by tweaking existing drivers. One thing I noticed is some of them are calling disp_wait_for_pending_transactions(), which shouldn't really ever be needed unless the specific hardware needs extra time between write calls or something. It may be worthwhile removing some of those calls to take out the extra waits where they are not needed. Calling disp_wait_for_pending_transactions() too often sort of undos the performance gains this new code adds.

C47D:
Hi,

Ran the widgets demo on my ILI9488 based display and I can replicate the results you had, in some screens the FPS are 33 with 3-4% CPU usage (the CPU is running at 240MHz). While other screens, the one with the graphs and gauges moving have 6 FPS and 100% CPU usage, this screen is kind of big (420x340) so it's a lot of data to send over the 40MHz SPI.

Will merge this and I will try to review the drivers which are calling `disp_wait_for_pending_transactions()`.

Thanks for the contribution @davidjade 👍

C47D (approved):

PR #210, opened 2020-10-06
Remove drivers symbols on lv_conf.

C47D:
This is to cleanup the `lv_conf` file from configuration depending on the chosen driver, the configuration is solved at `Kconfig` level.
All this work is towards using lvgl as a esp-idf component.

PR #214, opened 2020-10-08
README: Add table with supported display controllers and configurations.

C47D:
Let the users know they now will have to configure LVGL color depth and swap 16 colors in the lvgl menuconfig, it's no longer done automagically.

This is the follor up of #210

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,28 +12,28 @@ Monochrome support:
 
 ![Example_mono](images/new_mono.jpg)
 
-# Supported display controllers:
+# Supported display controllers
 
-## TFT
+**NOTE:** You now need to set the Color depth and Swap the 2 bytes of RGB565 color on the LVGL configuration menuconfig (it's no longer handled automatically).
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
Looks awesome!

kisvegabor (approved):

kisvegabor:
Maybe a max resolution filed and interface type (SPI, I2C, etc) could be useful. No pressure, just noting. :slightly_smiling_face:

embeddedt (approved):

C47D:
An interface field can be useful, I don't know about the resolution field, it can be anything the controller supports.

kisvegabor:
> I don't know about the resolution field, it can be anything the controller supports.

Actually the controllers have a max resolution. E.g. ILI9341 it's 320x240 so it can't be used for 480x272

C47D:
Will update the table later this week, maybe with a column named max horizontal resolution and max vertical resolution.

Issue #216, opened 2020-10-09
"Readable" LV_COLOR_DEPTH option in lv_conf.h


C47D:
Hi,

@embeddedt I think you can help me out on this one, I'm comparing the current `lv_conf.h` with the `lv_conf_internal.h`. This is because we want to handle all configurations inside `lv_conf_internal.h` so we don't need `lv_conf.h` to use LVGL as an ESP-IDF component.

Here's the issue:

Currently the `LV_COLOR_DEPTH` configuration looks like this in `lv_conf.h`:
```c
/* Color depth:
 * - 1:  1 byte per pixel
 * - 8:  RGB233
 * - 16: RGB565
 * - 32: ARGB8888
 */
#if defined (CONFIG_LV_COLOR_DEPTH_1)
#define LV_COLOR_DEPTH     1
#elif defined (CONFIG_LV_COLOR_DEPTH_8)
#define LV_COLOR_DEPTH     8
#elif defined (CONFIG_LV_COLOR_DEPTH_16)
#define LV_COLOR_DEPTH     16
#elif defined (CONFIG_LV_COLOR_DEPTH_32)
#define LV_COLOR_DEPTH     32
#endif
```

In `lv_conf_internal.h`:
```c
/* Color depth:
 * - 1:  1 byte per pixel
 * - 8:  RGB332
 * - 16: RGB565
 * - 32: ARGB8888
 */
#ifndef LV_COLOR_DEPTH
#  ifdef CONFIG_LV_COLOR_DEPTH
#    define LV_COLOR_DEPTH CONFIG_LV_COLOR_DEPTH
#  else
#    define  LV_COLOR_DEPTH     16
#  endif
#endif
```

To make them look alike we need to define CONFIG_LV_COLOR_DEPTH as an integer out of a Kconfig `choice` (which I don't think is possible), but applying [this](https://stackoverflow.com/a/31612706) workaround it is.

Applying the workaround to our Kconfig it looks like this:
```
choice LV_COLOR_DEPTH_BOOL
        prompt "Color depth."
        default LV_COLOR_DEPTH_16
        help
            Color depth to be used.

        config LV_COLOR_DEPTH_32
            bool "32: ARGB8888"
        config LV_COLOR_DEPTH_16
            bool "16: RGB565"
        config LV_COLOR_DEPTH_8
            bool "8: RGB232"
        config LV_COLOR_DEPTH_1
            bool "1: 1 byte per pixel"
    endchoice

[...]

PR #217, opened 2020-10-10
[LVGL] Refactor Kconfig and lv_conf.h

C47D:
Continuation from #216

### DIFF
--- a/components/lvgl/Kconfig
+++ b/components/lvgl/Kconfig
@@ -1,15 +1,36 @@
 menu "LVGL configuration"
 
-    config LV_COLOR_SCREEN_TRANSP
-        bool "Enable screen transparency."
+    choice
+        prompt "Screen transparency"
         help
             Useful for OSD or other overlapping GUIs.
             Requires `LV_COLOR_DEPTH = 32` colors and the screen's style
             should be modified: `style.body.opa = ...`.
+        default LV_COLOR_SCREEN_TRANSP_DISABLE
+        config LV_COLOR_SCREEN_TRANSP_DISABLE
+            "Disable LV_COLOR_SCREEN_TRANSP"
+        config LV_COLOR_SCREEN_TRANSP_ENABLE
+            "Enable LV_COLOR_SCREEN_TRANSP"
+    endchoice
+
+    config LV_COLOR_SCREEN_TRANSP
+        int
+        default 0 if LV_COLOR_SCREEN_TRANSP_DISABLE
+        default 1 if LV_COLOR_SCREEN_TRANSP_ENABLE
+
+    choice
+        prompt "Enable anti-aliasing (lines, and radiuses will be smoothed)."
+        default LV_ANTIALIAS_ENABLE
+        config LV_ANTIALIAS_DISABLE
+            bool "Disable LV_ANTIALIAS"
+        config LV_ANTIALIAS_ENABLE
+            bool "Enable LV_ANTIALIAS"
+    endchoice
 
     config LV_ANTIALIAS
-        bool "Enable anti-aliasing (lines, and radiuses will be smoothed)."
-        default y
+        int
+        default 0 if LV_ANTIALIAS_DISABLE
+        default 1 if LV_ANTIALIAS_ENABLE
 
     config LV_DISP_DEF_REFR_PERIOD
         int "Default display refresh period (ms)."
@@ -45,9 +66,19 @@ menu "LVGL configuration"
             displays fall in 4 categories. This is the upper limit for large
             displays.
 
+    choice
+        prompt "Swap the 2 bytes of RGB565 color. Useful if the display has a 8 bit interface (e.g. SPI)."
+        default LV_COLOR_16_SWAP_ENABLE
+        config LV_COLOR_16_SWAP_DISABLE
+            bool "Disable LV_COLOR_16_SWAP"
+        config LV_COLOR_16_SWAP_ENABLE
+            bool "Enable LV_COLOR_16_SWAP"
+    endchoice
+
     config LV_COLOR_16_SWAP
-        bool "Swap the 2 bytes of RGB565 color. Useful if the display has a 8 bit interface (e.g. SPI)."
-        default y
+        int
+        default 0 if LV_COLOR_16_SWAP_DISABLE
+        default 1 if LV_COLOR_16_SWAP_ENABLE
 
     choice LV_COLOR_DEPTH_BOOL
         prompt "Color depth."
--- a/components/lvgl/Kconfig
+++ b/components/lvgl/Kconfig
@@ -663,7 +663,7 @@ menu "LVGL configuration"
             bool "Enable the log module"
             default n
 
-        choice LV_LOG_LEVEL
+        choice LV_LOG_LEVEL_BOOL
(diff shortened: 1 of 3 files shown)


C47D:
@embeddedt I'm thinking on doing something similar to cde3024 for boolean symbols, what do you think?

embeddedt:
That looks like it would work.

C47D:
Ok, I will keep going and mark the PR as ready for review when I'm done.

kisvegabor (approved):
Looks good to me as well!

C47D:
I think I'm almost done, the bulk of the remaining work is on the fonts section of the `lv_conf.h` file. If the `CONFIG_LV_FONT_MONTSERRAT` symbol is not defined the font will be not enabled by the `lv_conf_internal.h`. So I think this is the expected behaviour.

`lv_conf_internal.h`
```c
#ifndef LV_FONT_MONTSERRAT_8
#  ifdef CONFIG_LV_FONT_MONTSERRAT_8
#    define LV_FONT_MONTSERRAT_8 CONFIG_LV_FONT_MONTSERRAT_8
#  else
#    define  LV_FONT_MONTSERRAT_8     0
#  endif
#endif
```

Other "issue" is how the fonts are setted in `lv_conf_ìnternal.h`:
```c
#ifndef LV_THEME_DEFAULT_FONT_NORMAL
#  ifdef CONFIG_LV_THEME_DEFAULT_FONT_NORMAL
#    define LV_THEME_DEFAULT_FONT_NORMAL CONFIG_LV_THEME_DEFAULT_FONT_NORMAL
#  else
#    define  LV_THEME_DEFAULT_FONT_NORMAL        &lv_font_montserrat_14
#  endif
#endif
```

`CONFIG_LV_THEME_DEFAULT_FONT_NORMAL` can´t be set to be something like `&lv_font_montserrat_14`, that is why we used something like this in the current `lv_conf.h`
```c
#if defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_12
#define LV_THEME_DEFAULT_FONT_NORMAL         &lv_font_montserrat_12
#elif defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_14
#define LV_THEME_DEFAULT_FONT_NORMAL         &lv_font_montserrat_14
#elif defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_16
#define LV_THEME_DEFAULT_FONT_NORMAL         &lv_font_montserrat_16
#elif defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_18
#define LV_THEME_DEFAULT_FONT_NORMAL         &lv_font_montserrat_18
#elif defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_20
/* same for all fonts */
```

So I think we should update the `lv_conf_internal.h` file with that approach.

Any opinions @embeddedt @kisvegabor ?

C47D:
Also I think we can split the Kconfig file into several (one for fonts, other for objects, etc.) using the `source` keyword (See [The Kernel Configuration and Build Process](https://www.linuxjournal.com/article/6568)).

kisvegabor:
> #if defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_12
#define LV_THEME_DEFAULT_FONT_NORMAL         &lv_font_montserrat_12
#elif defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_14
#define LV_THEME_DEFAULT_FONT_NORMAL         &lv_font_montserrat_14
#elif defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_16
#define LV_THEME_DEFAULT_FONT_NORMAL         &lv_font_montserrat_16
#elif defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_18
#define LV_THEME_DEFAULT_FONT_NORMAL         &lv_font_montserrat_18
#elif defined CONFIG_LV_FONT_DEFAULT_NORMAL_MONTSERRAT_20
/* same for all fonts */

It seems working but now `lv_conf_internal.h` is simply generated from `lv_conf_templ.h`. With this, we need to add some hardcoded logic to replace some parts.

I've opened a question about it in [Stackoverflow](https://stackoverflow.com/questions/64367658/select-a-variables-name-in-kconfig).

C47D:
I was able to load the Kconfig.fonts file into the main lvgl Kconfig file, so now we can split the different lvgl Kconfig menus into separated files, this would allow it to become more readable. Let me know what do you think, if you like that idea I will create a Kconfig file for the themes, log, debug and text settings and widgets.

C47D:
I think I'm wrong about the boolean configuration options because in the particular version of lvgl used in this project the `lv_conf_internal` doesn't handle undefined `CONFIG_x` options, that's why I have to always have the `CONFIG_x` symbols defined.

On the current lv_conf_internal (lvgl master) the non defined CONFIG_x options defaults to not enabled configurations.

kisvegabor:
> Let me know what do you think, if you like that idea I will create a Kconfig file for the themes, log, debug and text settings and widgets.

I like the idea in general. However, I'm afraid if we will have ~5-10 Kconfig files in the root folder it will look messy. IS there a way to put the Kconfig files into a folder?

> On the current lv_conf_internal (lvgl master) the non defined CONFIG_x options defaults to not enabled configurations.

Something like this would work better?
```c
#ifndef LV_ANTIALIAS  /*Only if it's defined somewhere...*/
/*If using with Kconfig*/
#  if LV_KCONFIG_ENV
#    ifdef CONFIG_LV_ANTIALIAS
#      define LV_ANTIALIAS   CONFIG_LV_ANTIALIAS
#    else
#      define LV_ANTIALIAS   0
#    endif
/*If not using Kconfig*/
#  else
#    define LV_ANTIALIAS   0
#  endif
#endif
```

Non-bool values are always defined, right?

C47D:
> I like the idea in general. However, I'm afraid if we will have ~5-10 Kconfig files in the root folder it will look messy. IS there a way to put the Kconfig files into a folder?

That's also true, I don't think is possible to put the Kconfig in other folder and make it work with both Zephyr and Esp-idf.

And about handling CONFIG_x symbols the current approach is fine.

I will open another PR and close this one.

embeddedt:
I think the `source` directive supports relative folder names, however, I don't know if they are interpreted relative to the `Kconfig` file's location, or to the current working directory.

C47D:
It does, I'm currently using a relative path, but taking the build directory of the esp-idf project. The Zephyr directory layout is different.

The rsource extension supports relative paths but starting from the Kconfig, so you only have to add `rsource Kconfig.fonts` and such.
The rsource is an extension by the Kconfiglib library that esp-idf uses to handle Kconfig files.

kisvegabor:
>  I'm currently using a relative path, but taking the build directory of the esp-idf project. The Zephyr directory layout is different.

If we need to hardcore project specific paths, I think, it's better to stick with monolithic Kconfig.

C47D:
> If we need to hardcore project specific paths, I think, it's better to stick with monolithic Kconfig.

I agree, in #219 I've keept the monolithic Kconfig and added some missing options, let's continue over there.

Issue #220, opened 2020-10-24
PlatformIO instructions are incomplete


cmumford:
I'm new to PlatformIO, but I'm pretty sure that the [platformio instructions](https://github.com/lvgl/lv_port_esp32#platformio-support) in README.md are incorrect. I read through issue #168 and it only recommends the following:

1. Start with [lv_platformio](https://github.com/lvgl/lv_platformio)
2. Modify `platformio.ini`
3. Modify `main.cc`

I think it is ambiguous whether it is necessary to create the git submodule in the lv_platformio project is required as described in [Build this repository standalone and run the demo](https://github.com/lvgl/lv_port_esp32#build-this-repository-standalone-and-run-the-demo). If this was done, then this would result in two copies of lvgl. One in `//components/lv_port_esp32/components/lvgl/lvgl/`, and a second in `//lib/lvgl/`. If the submodule is not created, then there is a missing step to add lv_port_esp32 as a dependency (lib_deps) in platformio.ini.

There is a [comment](https://github.com/lvgl/lv_port_esp32/issues/168#issuecomment-664386098) in issue 168 which says that following the README.md steps successfully builds - which it does, but the default app doesn't use any drivers and doesn't display anything.

Maybe somebody with more PlatformIO knowledge can see what I might be doing wrong, but from what I can see this project is not compatible (or ready to be used by) PlatformIO.

cmumford:
I managed to get my project running with lv_port_esp32. Here's what I had to do:

1. Checkout lv_port_esp32 into the //lib directory.
2. Make the following changes to lv_port_esp32:
   1. Include lvgl.h differently, PR #221 
   2. Include lv_refr.h diffently, PR #222.
   3. Added a library.json file (no PR yet - more on that later).
3. Made the following changes to my app's project:
   1. Modified my `platformio.ini` file to have "lv_port_esp32" in the "lib_deps" section.
   2. Copied `lib/lv_port_esp32/components/lvgl/lv_conf.` to `src/lv_conf.h`
   3. Copied the three Kconfig files from this project (skipping examples) to my project.
   4. Ran `platformio run -t menuconfig`, and set values as described by https://github.com/lvgl/lv_port_esp32#configuration-options.
   5. Added the LVGL code from main.c into my project's application.

Here is what I'd like to do to complete support for PlatformIO for this project - as far as I can tell lv_port_esp32 really isn't yet compatible with PlatformIO.

1. Land the two PR's above.
[...]

PR #221, opened 2020-10-25
Including "lvgl.h" if LV_LVGL_H_INCLUDE_SIMPLE is defined.

cmumford:
When LV_LVGL_H_INCLUDE_SIMPLE is not defined (the default) then
`"lvgl/lvgl.h"` is included, else `"lvgl.h"` is included.
This is for other build systems (like PlatformiIO) which setup
build paths differently than esp-idf. More info in issue #220.

This approach is identical to that used by https://github.com/lvgl/lv_drivers.


C47D:
Hi,

Thanks for the PR, this seems fine to me, I will read #220 tomorrow, I don't have a lot of experience using Platformio (only used it for a blinky project more than a year ago).

PR #223, opened 2020-10-26
Update DISPLAY_Y_OFFSET for TTGO predefined config

jonans:
Don't know if this differs between revisions but the Y offset is off by one on the TTGO T-DISPLAY predefined config.


C47D:
Thanks for the PR. I don't own that display, maybe the user that requested that offset field didn't noticed the offset being off by one pixel.

PR #226, opened 2020-10-29
Revert "Including lv_core/lv_refr.h differently form platformio."

cmumford:
This reverts commit 31c6ed0154600dd9ebe78f3538ed479be8b12aea.

The reverted change enabled a PlatformIO project to build lv_port_esp32
when adding it as a library dependency using the platformio.ini
lib_deps[¹] value. Unfortunately it broke the PlatformIO build when
building this project (i.e. main/main.c). Reverting and hopefully
coming up with a better solution.

[¹]: https://docs.platformio.org/en/latest/projectconf/section_env_library.html#lib-deps

### DIFF
--- a/components/lvgl_esp32_drivers/lvgl_helpers.c
+++ b/components/lvgl_esp32_drivers/lvgl_helpers.c
@@ -18,11 +18,7 @@
 
 #include "driver/i2c.h"
 
-#if defined(PLATFORMIO)
-#include "lv_core/lv_refr.h"
-#else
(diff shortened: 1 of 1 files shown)


C47D:
Hi @cmumford,

Thanks for the fix, sorry for not replying to the PR you just closed. We're near of having the LVGL main repo to behave like a ESP-IDF component by including the CMakeLists and Kconfig files directly to it, hopefully both files will be merged into the dev branch this weekend.

After that we can make this example repo more compatible with Platformio.

Regards

PR #234, opened 2020-12-08
Set overwrites main component

tlechman49:
Problem:
CMake cannot find the main component when it is configuring. This is cause by the set command overwriting the component list. Appending to the component list is a safer procedure.

Test Setup:
1. Initialize Platformio espidf blink example
2. Follow [these](https://github.com/lvgl/lv_port_esp32#install-this-project-as-a-library-submodule-in-your-own-project) steps from the readme file
3. Observe build errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ cmake_minimum_required(VERSION 3.5)
 
 include($ENV{IDF_PATH}/tools/cmake/project.cmake)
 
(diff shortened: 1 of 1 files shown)


C47D:
Thanks for the fix  @tlechman49, I'm not in my development machine but it seems fine for me.

PR #237, opened 2020-12-17
Drivers as submodule

C47D:
@kisvegabor This PR removes the `lvgl_esp32_drivers` directory and replace it with a submodule.


embeddedt (approved):

kisvegabor (approved):
Great, thanks!

Can we update the lvgl and lv_examples components too?

C47D:
I was planning to update one at the time. Because if we have any issues we can track down bugs more easily.

PR #265, opened 2021-02-24
Remove Extra component dirs definition

tore-espressif:
Latest `lvgl_esp32_drivers` changes its component structure. This project has to be updated for it.

TODOs:
- [x] Merge https://github.com/lvgl/lvgl_esp32_drivers/pull/42
- [x] Update lvgl_esp32_drivers submodule in this PR
- [x] Test on linux - Make
- [x] Test on linux - CMake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,7 @@ cmake_minimum_required(VERSION 3.5)
 
 if (NOT DEFINED PROJECT_NAME)
     include($ENV{IDF_PATH}/tools/cmake/project.cmake)
- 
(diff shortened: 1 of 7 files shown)


tore-espressif:
@C47D 
Would you know why this construct is used throughout the code, when we always use the simple_include variant? 

Is there a place somewhere where we don't use it?
``` C
#ifdef LV_LVGL_H_INCLUDE_SIMPLE
#include "src/lv_core/lv_refr.h"
#else
#include "lvgl/src/lv_core/lv_refr.h"
#endif
```

C47D:
@tore-espressif AFAIK, it's used so the users can or not enable the simple_include variant on their projects.

PR #287, opened 2021-07-16
Add esp-build action

tore-espressif:
This is a preview how to run ESP-IDF build from Github Actions.

It's a 'preview' because there are few limitations:

- ~~We still need `lv_port_esp32` repo to run build action, because `lvgl_esp32_drivers` doesn't contain examples that can be built into a full app~~ There are examples and build action now in `lvgl_esp32_drivers`
- ~~Currently only default example configuration is built. Thus only ILI9341 driver is compiled. Ultimately we need to compile all drivers in one build action~~. `master` of `lvgl_esp32_drivers` builds all drivers

Above-mentioned issues will be addressed in https://github.com/lvgl/lvgl_esp32_drivers/issues/72

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,29 @@
+name: 'build'
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        idf_ver: ["v4.1", "v4.2", "v4.3"]
+        idf_target: ["esp32"]
+        include:
+          - idf_ver: "v4.2"
+            idf_target: esp32s2
+            # TODO: enable C3 build
+          #- idf_ver: "v4.3"
+          #  idf_target: esp32c3
+    runs-on: ubuntu-20.04
+    container: espressif/idf:release-${{ matrix.idf_ver }}
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          submodules: recursive
+      - name: Build for ESP32
+        env:
+          IDF_TARGET: ${{ matrix.idf_target }}
+        shell: bash
+        run: |
+          . ${IDF_PATH}/export.sh
+          idf.py build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
     steps:
       - uses: actions/checkout@v1
         with:
-          submodules: recursive
+          submodules: true
       - name: Build for ESP32
         env:
           IDF_TARGET: ${{ matrix.idf_target }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ jobs:
             # TODO: enable C3 build
           #- idf_ver: "v4.3"
           #  idf_target: esp32c3
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     container: espressif/idf:release-${{ matrix.idf_ver }}
     steps:
       - uses: actions/checkout@v1


tore-espressif:
ESP32-C3 build is commented out for now, as the C3 support is not implemented yet.

kisvegabor (approved):
Great, looks good to me!

kisvegabor:
BTW, is there at least one driver that is built during the test?

tore-espressif:
only https://github.com/lvgl/lvgl_esp32_drivers/blob/master/lvgl_tft/ili9341.c
which is selected by default

kisvegabor:
IMO it might be useful to test all the drivers but it's more important to get started with CI.

tore-espressif:
Exactly, as said in the PR description, this is just a teaser trailer for anyone interested in Github Actions. (I'm new to it too, but it's rather similar to Gitlab CI).

I'd like to change the way we build and link esp32_lvgl drivers (build them all) rather then creating 20+ runs of this pipeline :D

embeddedt:
> We still need lv_port_esp32 repo to run build action, because lvgl_esp32_drivers doesn't contain examples that can be built into a full app

With `lv_micropython`, we have a similar setup: a main repository with the build system, and a submodule with drivers/bindings.

The [CI workflow on the bindings](https://github.com/lvgl/lv_binding_micropython/blob/master/.github/workflows/unix_port.yml) clones the main repository and then checks out the required bindings commit. This is not perfect, and will always produce errors when there is API breakage, but it works 90% of the time.

I think a similar solution could be adopted for the ESP32 repository.

> I'd like to change the way we build and link esp32_lvgl drivers (to build them all) rather then creating 20+ runs of this pipeline :D

Keep in mind that (at the moment) public GitHub repositories have unlimited Actions minutes, so we don't need to be *too* efficient when it comes to the number of runs, if it's easier to build individual drivers.

.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
...
+        include:
+          - idf_ver: "v4.2"
+            idf_target: esp32s2
+            # TODO: enable C3 build
+          #- idf_ver: "v4.3"
+          #  idf_target: esp32c3
+    runs-on: ubuntu-20.04
+    container: espressif/idf:release-${{ matrix.idf_ver }}
+    steps:
+      - uses: actions/checkout@v1
```
embeddedt, line 20: Checkout v2 does shallow checkouts by default, which makes it a lot faster, and probably reduces the bandwidth costs on GitHub's end as well. 

```suggestion
      - uses: actions/checkout@v2
```

tore-espressif: The action fails with checkout@v2, will investigate further
**EDIT:** Seems like we don't have `git v2.18` in our esp-idf container?

embeddedt: It's [supposed to have `git`](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/tools/idf-docker-image.html), however, I wouldn't be surprised if the version is older than what `actions/checkout@v2` expects.

C47D:
@tore-espressif Thanks for getting this work started, I'm also very interested to know how could we build all (or most) of the drivers.

Should we consider refactoring how the drivers are being build? currently we only make available one display (and one indev) driver per project, but I've seen interest in being able to build a project with multiple display drivers. Should we add that possibility to this?

.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
...
+          #  idf_target: esp32c3
+    runs-on: ubuntu-20.04
+    container: espressif/idf:release-${{ matrix.idf_ver }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Build for ESP32
+        env:
+          IDF_TARGET: ${{ matrix.idf_target }}
```
C47D, line 25: @tore-espressif I have a silly question, if we add more targets into the `idf_target` "vector" (can we call it vector?) the job will run on all the combinations between `idf_ver` and `idf_target`?

tore-espressif: Exactly, as it is now, it would run 3x for ESP32 (IDF v 4.1, 4.2 and 4.3) and once for ESP32-S2 (v4.2). So adding new IDF versions/target should be intuitive.

I removed version 4.0 and 3.3 because it would fail on `rsource` in Kconfig, which is supported from v4.1 onwards. https://github.com/lvgl/lvgl_esp32_drivers/blob/9fed1cc47b5a45fec6bae08b55d2147d3b50260c/Kconfig#L1.

C47D: Thanks for the reply, let us know when you want to merge this.

.github/workflows/build.yml
tore-espressif: I'll revert to checkout@v1 for now

tore-espressif:
https://github.com/lvgl/lvgl_esp32_drivers/pull/96 builds all display drivers.

> The CI workflow on the bindings clones the main repository and then checks out the required bindings commit. This is not perfect, and will always produce errors when there is API breakage, but it works 90% of the time.

I'll check this option too

.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
...
+            idf_target: esp32s2
+            # TODO: enable C3 build
+          #- idf_ver: "v4.3"
+          #  idf_target: esp32c3
+    runs-on: ubuntu-latest
+    container: espressif/idf:release-${{ matrix.idf_ver }}
+    steps:
+      - uses: actions/checkout@v1
+        with:
+          submodules: recursive
```
tore-espressif: Try setting this to `true` instead of `recursive`

C47D:
Is this ready to be merged?

tore-espressif:
Yes, it is ready!

Some of the previously mentioned limitations are already solved in `master` of `lvgl_esp32_drivers` :

- ESP32C3 is now supported
- All drivers are build, ensuring their correctness

C47D:
Great, you're really making awesome changes :smile

PR #301, opened 2022-01-11
Update main.c

Coffeeye:
Call functions with params for porting v8. (Legacy NULLS from v7 deleted)

Wait correct amount of time in while loop. Use assets given by functions.

issue : https://github.com/lvgl/lvgl/issues/1839#issuecomment-705440508

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -147,12 +147,11 @@ static void guiTask(void *pvParameter) {
     create_demo_application();
 
     while (1) {
-        /* Delay 1 tick (assumes FreeRTOS tick is 10ms */
-        vTaskDelay(pdMS_TO_TICKS(10));
-
         /* Try to take the semaphore, call lvgl related function on success */
         if (pdTRUE == xSemaphoreTake(xGuiSemaphore, portMAX_DELAY)) {
-            lv_task_handler();
+            uint32_t ticks_to_wait = lv_task_handler();
+            if(ticks_to_wait == 0) continue;
+            vTaskDelay(ticks_to_wait);
             xSemaphoreGive(xGuiSemaphore);
        }
     }
@@ -177,7 +176,7 @@ static void create_demo_application(void)
     lv_obj_t * scr = lv_disp_get_scr_act(NULL);
 
     /*Create a Label on the currently active screen*/
-    lv_obj_t * label1 =  lv_label_create(scr, NULL);
+    lv_obj_t * label1 =  lv_label_create(scr);
 
     /*Modify the Label's text*/
     lv_label_set_text(label1, "Hello\nworld");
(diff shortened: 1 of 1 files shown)


main/main.c
```diff
@@ -147,12 +147,11 @@ static void guiTask(void *pvParameter) {
...
 
     while (1) {
-        /* Delay 1 tick (assumes FreeRTOS tick is 10ms */
-        vTaskDelay(pdMS_TO_TICKS(10));
-
         /* Try to take the semaphore, call lvgl related function on success */
         if (pdTRUE == xSemaphoreTake(xGuiSemaphore, portMAX_DELAY)) {
-            lv_task_handler();
+            uint32_t ticks_to_wait = lv_task_handler();
+            if(ticks_to_wait == 0) continue;
```
Coffeeye, line 153: Should be considered as '''unwanted''' due to prevent mutex release.

Sorry

C47D:
Thanks for the PR @Coffeeye , I'm still not sure how are we going to handle support for multiple versions of LVGL on `main.c`. Maybe adding `#if` guards might be a possible idea? such as

```c
#if LVGL_VERSION_MAJOR >= 8
        /* LVGL v8 */
#else
        /* LVGL v7 */
#endif
```

Any ideas on how to proceed @tore-espressif ?

Coffeeye:
This is a fork of the project named porting v8 so imo .md should be only with that info and examples made typical for v8.

If I were you @C47D , that would be beyond doubt

main/main.c
Coffeeye: I discovered also that using function return with lower than default HZ rate in FreeRTOS (which is 1000) makes display very "laggy". Which makes sense but maybe some1 will be using this like this and experience that symptom.

C47D:
Yup, the port_v8 branch is pretty old and it was just an experiment. The current intention is to support both versions of LVGL on the same example.

main/main.c
sukesh-ak: > I discovered also that using function return with lower than default HZ rate in FreeRTOS (which is 1000) makes display very "laggy". Which makes sense but maybe some1 will be using this like this and experience that symptom.

What did you mean by this? I think I am having this issue of lag.

sukesh-ak:
> Yup, the port_v8 branch is pretty old and it was just an experiment. The current intention is to support both versions of LVGL on the same example.

Finally found LovyanGFX the best drivers which works with v8.x for both Arduino and ESP-IDF.
ESP-IDF & Arduino samples for WT32-SC01 using LVGL8 + LOVYANGFX 

Device : WT32-SC01 
Display / Touch Drivers : LovyanGFX
Widgets : LVGL 8.x

ESP-IDF => WT32-SC01 + LVGL8 + LOVYANGFX 
https://github.com/sukesh-ak/LVGL8-WT32-SC01-IDF

Arduino => WT32-SC01 + LVGL8 + LOVYANGFX 
https://github.com/sukesh-ak/LVGL8-WT32-SC01-Arduino

PR #303, opened 2022-03-04
Use FreeRTOS instead of ESP timer

Coffeeye:
I understand this is esp_port but using esp timer causes buffer overflow which leads to panic using built in ESP-IDF modules. 
My suggestion - use the timer from the RTOS system thats already used in the example and also include placing the call in IRAM to increase its responsiveness.

### DIFF
--- a/main/main.c
+++ b/main/main.c
@@ -52,7 +52,7 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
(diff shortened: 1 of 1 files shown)


mkfrey:
Why does using esp_timer cause a buffer overflow? If a stack overflow happens, you can increase the timer task size.

